PR #11, opened 2023-08-23
Add Earthfile and GHA to run on PR + merge

expressvpn-ian-c:
Basically a simple lift-and-shift from the wolfssl-sys CI, with minor tweaks to the files copied into the build containers.

I also had to use Rust 1.71 since 1.69 does not build this crate (we use newer features).


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    branches: [main]
+  pull_request: null # target every PR
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: earthly/actions/setup-earthly@v1
+        with:
+          version: v0.6.30
```
expressvpn-mariappan-r, line 12:
```suggestion
      - uses: earthly/actions-setup@v1
        with:
          version: v0.7.15
```
The `actions` repo has been moved to `actions-setup` 
https://github.com/earthly/actions

Also think, we can use the latest earthly version

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v2
```
expressvpn-mariappan-r, line 13:
```suggestion
      - uses: actions/checkout@v3
```
Also v3 available

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
```
expressvpn-mariappan-r, line 2: Curious why are we ignoring the lock file ?

Earthfile
```diff
@@ -0,0 +1,44 @@
+VERSION 0.6
```
expressvpn-mariappan-r, line 1:
```suggestion
VERSION 0.7
```

.github/workflows/ci.yaml
expressvpn-ian-c: Ack, I'll do this and the v3 of the checkout action..

I think I'll also now raise a parallel PR against https://github.com/expressvpn/wolfssl-sys to refresh that too

Earthfile
```diff
@@ -0,0 +1,44 @@
...
+
+WORKDIR /wolfssl-sys
+
+build-deps:
+    RUN apt-get update -qq
+    RUN apt-get install --no-install-recommends -qq autoconf autotools-dev libtool-bin clang cmake
+    RUN apt-get -y install --no-install-recommends bsdmainutils
+    RUN rustup component add rustfmt
+
+copy-src:
```
expressvpn-mariappan-r, line 12: FYI, I have seen some example in earthly docs to cache dependencies efficiently using cargo-chef:

https://github.com/earthly/earthly/blob/main/examples/rust/Earthfile#L14

We can look into it later too.

.gitignore
expressvpn-ian-c: IMHO we should not be. I'm going to commit a current one and ensure that we use it for CI.

Earthfile
expressvpn-ian-c: Yes, I've used cargo chef before. It can be a bit fickle to get it to work reliably, but we should certainly consider it for future work.

.github/workflows/ci.yaml
expressvpn-ian-c: Parallel PR: https://github.com/expressvpn/wolfssl-sys/pull/38

kp-mariappan-ramasamy (approved):
LGTM

PR #14, opened 2023-08-25
Remove I/O buffering from the bindings layer.

expressvpn-ian-c:
From the penultimate commit message:

> Buffering in the bindings breaks the link between the actual I/O (done by the consuming application) and the WolfSSL state machine, making it hard for the consumer to know when it needs to process I/O and whether that I/O will be sending or receiving and meaning that WolfSSL does not have an acurate view of the state of the I/O (since the bindings say it has been done when it has been buffered).

By fixing the I/O model we can remove the spurious run_io_loop from the integration tests.

I've avoided `Box<dyn IOCallbacks>` here in favor of plumbing a `IOCB: IOCallbacks` generic type through the `ssl::Session` code so we can have `Box<IOCallbacks>`. Normally I would favour the simplicity of `dyn IOCallbacks` however that implies dynamic dispatch and this is code is on the data path where we'd rather not pay that price[^0]. Since we use a generic type the `Session::io_*_shim` implementations are monomorphized into a direct call to the application provided method:

```console
$ cargo build --test ping_pong --release
$ objdump -CSd target/release/deps/ping_pong-73aae058390bb251 | grep -A 13 io_recv_shim\>:
000000000001bb20 <wolfssl::ssl::Session<IOCB>::io_recv_shim>:
   1bb20:	55                   	push   %rbp
   1bb21:	41 56                	push   %r14
   1bb23:	53                   	push   %rbx
   1bb24:	48 83 ec 10          	sub    $0x10,%rsp
   1bb28:	48 89 f0             	mov    %rsi,%rax
   1bb2b:	4c 63 c2             	movslq %edx,%r8
   1bb2e:	48 89 e7             	mov    %rsp,%rdi
   1bb31:	48 89 ce             	mov    %rcx,%rsi
   1bb34:	48 89 c2             	mov    %rax,%rdx
   1bb37:	4c 89 c1             	mov    %r8,%rcx
   1bb3a:	e8 d1 a6 00 00       	call   26210 <<ping_pong::SockIOCallbacks<IOCB> as wolfssl::callback::IOCallbacks>::recv>
   1bb3f:	48 8b 04 24          	mov    (%rsp),%rax
   1bb43:	48 85 c0             	test   %rax,%rax
... Snip the conversion from Rust error to WolfSSL error types
```

Note the direct call to `<<ping_pong::SockIOCallbacks<IOCB> as wolfssl::callback::IOCallbacks>::recv>`. I think that's as good as it gets without imposing unsafety on the application.

There is a downside to this monomorphisation since it means that one such shim is generated for each `impl IOCallbacks for` (e.g. `target/release/deps/ping_pong-73aae058390bb251` actually has two -- one for the dtls test case and one for the tls test case). It my assumption that any given app will have a small number of IO paths (I suspect the mode will be 1).

[^0]: This goes against my usual mantra of "optimize based on profiling" so could be considered a premature optimization, but I think in this context that's ok, the complexity of the generic types isn't too intolerable....


expressvpn-pete-m (approved):
Really like the ping/pong tests :slightly_smiling_face: 

Given these read / write calls are on the critical path for processing data, and we want to offer good performance for hardware implementations that might not be well optimised for dynamic dispatch, I wouldn't class this as premature optimisation :slightly_smiling_face:

PR #15, opened 2023-08-30
Add cargo deny to audit licenses of dependencies

expressvpn-mariappan-r:
**Changes:**
- Add deny.toml which `cargo-deny` uses to check license
- Add earthly target `check-license` for checking licenses
- Update github actions to run this target on all PRs

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,53 @@
+[advisories]
+db-path = "~/.cargo/advisory-db"
+db-urls = ["https://github.com/rustsec/advisory-db"]
+vulnerability = "deny"
+unmaintained = "deny"
+yanked = "deny"
+notice = "deny"
+ignore = [
+    #"RUSTSEC-0000-0000",
+]
+# * None - CVSS Score 0.0
+# * Low - CVSS Score 0.1 - 3.9
+# * Medium - CVSS Score 4.0 - 6.9
+# * High - CVSS Score 7.0 - 8.9
+# * Critical - CVSS Score 9.0 - 10.0
+severity-threshold = "High"
+
+[licenses]
+default = "deny"
+copyleft = "deny"
+unlicensed = "deny"
+allow = [
+    "MIT",
+    "Apache-2.0",
+    "Unicode-DFS-2016",
+    "BSD-3-Clause",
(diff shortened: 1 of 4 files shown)


deny.toml
```diff
@@ -0,0 +1,272 @@
...
+# List of explicitly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+allow = [
+    "MIT",
+    "Apache-2.0",
+    "GPL-2.0",
+    "Unicode-DFS-2016",
+    "BSD-3-Clause",
+    "ISC",
```
expressvpn-mariappan-r, line 110: Please check whether this licenses are fine to allow

deny.toml
```diff
@@ -0,0 +1,272 @@
...
+# if not specified. If it is specified but empty, no registries are allowed.
+allow-registry = ["https://github.com/rust-lang/crates.io-index"]
+# List of URLs for allowed Git repositories
+allow-git = [
+  "https://github.com/open-quantum-safe/liboqs-rust",
+]
+
+[sources.allow-org]
+# 1 or more github.com organizations to allow git sources for
+github = ["expressvpn"]
```
expressvpn-mariappan-r, line 272: Allowed `expressvpn` org and `liboqs-rust` repo

expressvpn-ian-c: If possible we should switch back to the releases liboqs-rust and avoid that one.

For the expressvpn/wolfssl-sys dep I think we should explicitly allow that git repo rather than the whole orga, IMHO we should consciously decide to rely on specific git repositories.

expressvpn-ian-c (commented):
One specific comment here, but the general comments from https://github.com/expressvpn/wolfssl-sys/pull/40 mostly apply too.

deny.toml
expressvpn-mariappan-r: Not able to move to crate version now as explained here: https://github.com/expressvpn/wolfssl-sys/pull/40#discussion_r1310476848

Yes, agreed on specifying git repos explicitly

expressvpn-ian-c (approved):

PR #23, opened 2023-09-07
test satellites

idelvall:
This PR:
- Collapses the Earthly targets run on `ci.yaml` into a new `+ci` target, so only one GH runner is needed for them
- Makes use of https://github.com/earthly/lib/blob/main/rust/Earthfile#L10 to cache cargo dependency downloads and compilations across builds
- Runs earthly in a dedicated satellite, so cache is maintained across different GH action runs

As a result, this PR reduces GH actions build time from around 8m to 50s for the most likely scenario of only source code changing

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,50 +6,18 @@ on:
 
 jobs:
   ci:
-    needs: [test, build, fmt, lint, cargo-deny]
     runs-on: ubuntu-latest
-    steps:
-      - shell: bash
-        run: |
-          echo "Build success"
-  test:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: earthly/actions-setup@v1
-        with:
-          version: v0.7.15
-      - uses: actions/checkout@v3
-      - name: Run Tests
-        run: earthly --ci +run-tests
-  build:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: earthly/actions-setup@v1
-        with:
-          version: v0.7.15
-      - uses: actions/checkout@v3
-      - name: Build crate
-        run: earthly --ci +build-release
-  fmt:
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1,5 +1,5 @@
 //! The `wolfssl` crate is designed to be a Rust layer built on top of
-//! the `wolfssl-sys` crate (a C passthrough crate).
+//! the `wolfssl-s   ys` crate (a C passthrough crate).
```
expressvpn-brendan-h, line 2: Typo here? Extra space inserted in crate name

idelvall: Hey thanks Brendan, 
I made that change while in draft mode, to trigger a new build and check caching was working properly and the changes were properly compiled.
Rolled it back before marking the PR ready for review.

idelvall:
Refactor the code to use the new version github.com/earthly/lib/rust:2.1.0, that simplifies it's interface and adds caching to the cargo git metadata. 
This brings the build time when sources change down to 50s 🦾

idelvall:
squashed all commits

idelvall:
Squashed all commits, this should be ready.
I would appreciate a review. Thanks guys!

idelvall:
Just moved `cargo-deny` also to run in Earthly. 
@expressvpn-mariappan-r please check if this is aligned with you latest changes (i think it is 🙂), thanks!

expressvpn-mariappan-r:
Thanks @idelvall, Changes looks good to me. 
Thanks to @expressvpn-ian-c 's suggestion, we renamed the job name to "ci" inline to your pr :partying_face:

expressvpn-andre-l:
Thanks again to @idelvall for raising this PR!  I've rebased the changes in https://github.com/expressvpn/wolfssl-rs/pull/44, so I will close this PR now.

PR #24, opened 2023-09-08
Enable clippy undocumented_unsafe_blocks lint

expressvpn-mariappan-r:
**Code changes:**
- Enable clippy undocumented_unsafe_blocks lint in CI
- As first step, add `#[allow(undocumented_unsafe_blocks)]` for all unsafe code
- Added safety comments for following files:
   - lib.rs
   - context.rs
   - error.rs
   - rng.rs
   -  ssl.rs

**Note:**
There are two block left without SAFETY comments in `wolf_init` and `wolf_cleanup`   for now
Will fix up in followup pr

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,5 +1,3 @@
-#![allow(clippy::undocumented_unsafe_blocks)]
-
 use crate::{
     callback::IOCallbacks,
     error::{Error, Result},
@@ -37,6 +35,10 @@ impl ContextBuilder {
             .into_method_ptr()
             .ok_or(NewContextBuilderError::MethodFailed)?;
 
+        // SAFETY: [`wolfSSL_CTX_new`][0] is documented to get pointer to a valid `WOLFSSL_METHOD` structure which is created using one of the `wolfSSLvXX_XXXX_method()`.
+        // `Protocol::into_method_ptr` function returns a pointer `wolfSSLvXX_XXXX_method()`
+        //
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfssl_ctx_new
         let ctx = unsafe { wolfssl_sys::wolfSSL_CTX_new(method_fn.as_ptr()) };
         let ctx = NonNull::new(ctx).ok_or(NewContextBuilderError::CreateFailed)?;
 
@@ -54,6 +56,12 @@ impl ContextBuilder {
         };
 
         let result = match root {
+            // SAFETY: [`wolfSSL_CTX_load_verify_buffer`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+            // The pointer given as the `in` argument must point to a region of `sz` bytes.
+            // The values passed here are valid since they are derived from the same byte slice.
+            //
+            // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_load_verify_buffer
+            // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gaa5a28f0ac25d9abeb72fcee81bbf647b
             RootCertificate::Asn1Buffer(buf) => unsafe {
                 wolfSSL_CTX_load_verify_buffer(
                     self.ctx.as_ptr(),
@@ -62,6 +70,12 @@ impl ContextBuilder {
                     WOLFSSL_FILETYPE_ASN1,
                 )
             },
+            // SAFETY: [`wolfSSL_CTX_load_verify_buffer`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+            // The pointer given as the `in` argument must point to a region of `sz` bytes.
+            // The values passed here are valid since they are derived from the same byte slice.
+            //
+            // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_load_verify_buffer
+            // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gaa5a28f0ac25d9abeb72fcee81bbf647b
             RootCertificate::PemBuffer(buf) => unsafe {
                 wolfSSL_CTX_load_verify_buffer(
                     self.ctx.as_ptr(),
@@ -78,6 +92,12 @@ impl ContextBuilder {
                 let path = std::ffi::CString::new(path)
                     .or(Err(Error::fatal(wolfssl_sys::WOLFSSL_BAD_PATH)))?;
                 if is_dir {
+                    // SAFETY: [`wolfSSL_CTX_load_verify_locations`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+                    // If not NULL, then the pointer passed as the path argument must be a valid NULL-terminated C-style string,
+                    // which is guaranteed by the use of `std::ffi::CString::as_c_str()` here.
+                    //
+                    // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_load_verify_locations
+                    // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gaf592c652b5d7a599ee511a394dfc488e
                     unsafe {
                         wolfSSL_CTX_load_verify_locations(
                             self.ctx.as_ptr(),
@@ -86,6 +106,12 @@ impl ContextBuilder {
                         )
                     }
                 } else {
+                    // SAFETY: [`wolfSSL_CTX_load_verify_locations`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+                    // If not NULL, then the pointer passed as the path argument must be a valid NULL-terminated C-style string,
+                    // which is guaranteed by the use of `std::ffi::CString::as_c_str()` here.
+                    //
+                    // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_load_verify_locations
+                    // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gaf592c652b5d7a599ee511a394dfc488e
                     unsafe {
                         wolfSSL_CTX_load_verify_locations(
                             self.ctx.as_ptr(),
@@ -111,6 +137,12 @@ impl ContextBuilder {
         let cipher_list = std::ffi::CString::new(cipher_list)
             .or(Err(Error::fatal(wolfssl_sys::WOLFSSL_FAILURE)))?;
 
+        // SAFETY: [`wolfSSL_CTX_set_cipher_list`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()` and
+        // `list` parameter which should be a null terminated C string pointer which is guaranteed by
+        // the use of `std::ffi::CString::as_c_str()` here.
+        //
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/ssl_8h.html#function-wolfssl_ctx_set_cipher_list
+        // [1]: https://www.wolfssl.com/doxygen/group__Setup.html#gafa55814f56bd7a36f4035d71b2b31832
         let result = unsafe {
             wolfssl_sys::wolfSSL_CTX_set_cipher_list(
                 self.ctx.as_ptr(),
@@ -136,6 +168,12 @@ impl ContextBuilder {
         };
 
         let result = match secret {
+            // SAFETY: [`wolfSSL_CTX_use_certificate_buffer`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+            // The pointer given as the `in` argument must point to a region of `sz` bytes.
+            // The values passed here are valid since they are derived from the same byte slice.
+            //
+            // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_use_certificate_buffer
+            // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gae424b3a63756ab805de5c43b67f4df4f
             Secret::Asn1Buffer(buf) => unsafe {
                 wolfSSL_CTX_use_certificate_buffer(
                     self.ctx.as_ptr(),
@@ -144,18 +182,32 @@ impl ContextBuilder {
                     WOLFSSL_FILETYPE_ASN1,
                 )
             },
-            Secret::Asn1File(path) => unsafe {
+            Secret::Asn1File(path) => {
                 let path = path
                     .to_str()
                     .ok_or(Error::fatal(wolfssl_sys::BAD_PATH_ERROR))?;
                 let file = std::ffi::CString::new(path)
                     .or(Err(Error::fatal(wolfssl_sys::BAD_PATH_ERROR)))?;
-                wolfSSL_CTX_use_certificate_file(
-                    self.ctx.as_ptr(),
-                    file.as_c_str().as_ptr(),
-                    WOLFSSL_FILETYPE_ASN1,
-                )
-            },
+                // SAFETY: [`wolfSSL_CTX_use_certificate_file`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+                // The pointer passed as the path argument must be a valid NULL-terminated C-style string,
+                // which is guaranteed by the use of `std::ffi::CString::as_c_str()` here.
+                //
+                // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_use_certificate_file
+                // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#ga5a31292b75b4caa4462a3305d2615beb
+                unsafe {
+                    wolfSSL_CTX_use_certificate_file(
+                        self.ctx.as_ptr(),
+                        file.as_c_str().as_ptr(),
+                        WOLFSSL_FILETYPE_ASN1,
+                    )
+                }
+            }
+            // SAFETY: [`wolfSSL_CTX_use_certificate_buffer`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.
+            // The pointer given as the `in` argument must point to a region of `sz` bytes.
+            // The values passed here are valid since they are derived from the same byte slice.
+            //
+            // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__CertsKeys.html#function-wolfssl_ctx_use_certificate_buffer
+            // [1]: https://www.wolfssl.com/doxygen/group__CertsKeys.html#gae424b3a63756ab805de5c43b67f4df4f
             Secret::PemBuffer(buf) => unsafe {
                 wolfSSL_CTX_use_certificate_buffer(
                     self.ctx.as_ptr(),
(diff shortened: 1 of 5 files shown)


Earthfile
```diff
@@ -40,7 +40,7 @@ build-crate:
 lint:
     FROM +copy-src
     RUN rustup component add clippy
-    RUN cargo clippy --all-features --all-targets -- -D warnings
+    RUN cargo clippy --all-features --all-targets -- -D warnings -D clippy::undocumented_unsafe_blocks
```
expressvpn-ian-c, line 43: Lets instead do this via `#![warn( clippy::undocumented_unsafe_blocks)]` at the top of `lib.rs`, so devs can just run `cargo clippy` .

src/error.rs
```diff
@@ -94,6 +94,12 @@ pub type Result<T> = std::result::Result<T, Error>;
 // `wolfSSL_get_error`, which returns a `c_int`
 fn wolf_error_string(raw_err: std::ffi::c_ulong) -> String {
     let mut buffer = vec![0u8; wolfssl_sys::WOLFSSL_MAX_ERROR_SZ as usize];
+    // SAFETY:
+    // The underlying `C` api copies the static c string using strncpy with maximum s↴ize `WOLFSSL_MAX_ERROR_SZ`
+    // So we create a `mutable` buffer with size `WOLFSSL_MAX_ERROR_SZ` to satify the api req
+    // https://github.com/wolfSSL/wolfssl/blob/v5.6.3-stable/src/ssl.c#L4560
+    // https://github.com/wolfSSL/wolfssl/blob/v5.6.3-stable/src/internal.c#L23628
```
expressvpn-ian-c, line 101: I think the first/primary focus of the safety comments should be the wolfssl documentation.

The comments should only need to delve into references to specific code if the documentation is lacking/imprecise/ambiguous.

The wolfSSL people are IMHO at liberty to change the implementation in arbitrary ways, but the API docs are their contract with us.

So here I'd expect primarily to reference: https://www.wolfssl.com/doxygen/group__Debug.html#ga91d8474ba8abcf3fe594928056834993 and/or https://www.wolfssl.com/documentation/manuals/wolfssl/group__Debug.html#function-wolfssl_err_error_string not the code.
e.g.

```rust
/// SAFETY:
/// [`wolfSSL_ERR_error_string()`][0] is documented to store at most `WOLFSSL_MAX_ERROR_SZ` bytes, so `buffer` is appropriately sized.
///
/// [0]: https://www.wolfssl.com/doxygen/group__Debug.html#ga91d8474ba8abcf3fe594928056834993
```

(An aside on the implementation: On failure a static string is returned describing the problem -- we don't deal with that case here, probably we should).

expressvpn-ian-c (commented):
LGTM, I had a lot of nits but about half of them were about string slice vs byte slice...

There's one or two places where I've strayed into comments on the code itself rather than the safety comments, in the hopes of making things safer overall. Arguably they are a little out of scope, sorry. I justify it since the changes are related to making the comments true/accurate.

src/lib.rs
```diff
@@ -38,7 +38,10 @@ const TLS_MAX_RECORD_SIZE: usize = 2usize.pow(14) + 1;
 ///
 /// [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__TLS.html#function-wolfssl_init
 pub fn wolf_init() -> Result<()> {
-    #[allow(clippy::undocumented_unsafe_blocks)]
+    // SAFETY: [`wolfSSL_Init`][0][1] is documented to be called only once per application before calling other api's
```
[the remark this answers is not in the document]
expressvpn-mariappan-r: Yes, `std::sync::Once` sounds good, I will update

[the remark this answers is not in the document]
expressvpn-mariappan-r: This function returns `wolfssl_init `api error. Not sure how to do with Once.
Will fix this and the below `wolfssl_cleanup` api in followup PR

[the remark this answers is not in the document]
expressvpn-ian-c: https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html can help?

`once_cell` has been partially moved into `std` (that's where `std::sync::Once` comes from), but there are some bits which are still to come, I believe the intention is for the Lazy stuff to become part of `std` eventually so it's a good choice.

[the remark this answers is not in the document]
expressvpn-ian-c: FYI https://github.com/rust-lang/rust/issues/109736 is the tracking issue...

src/context.rs
```diff
@@ -39,6 +40,10 @@ impl ContextBuilder {
         };
 
         let result = match root {
+            // SAFETY: [`wolfSSL_CTX_load_verify_buffer`][0][1] takes buffer with size specified as `sz`.
+            // `buf` is valid, since it is from a string slice
```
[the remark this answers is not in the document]
expressvpn-mariappan-r: Done

src/context.rs
```diff
@@ -22,6 +20,9 @@ impl ContextBuilder {
     pub fn new(protocol: Protocol) -> Option<Self> {
         let method_fn = protocol.into_method_ptr()?;
 
+        // SAFETY: [`wolfSSL_CTX_new`][0] is documented to get pointer to the valid `WOLFSSL_METHOD` structure which is created using one of the wolfSSLvXX_XXXX_method().
+        // `Protocol`'s into_method_ptr function returns pointer  wolfSSLvXX_XXXX_method()
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfssl_ctx_new
```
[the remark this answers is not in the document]
expressvpn-mariappan-r: Thanks, address all comments.
Added blank line before link ref in all comments.
Yes, `Protocol::into_method_ptr` looks more idiomatic and clear

src/error.rs
```diff
@@ -95,6 +94,12 @@ pub type Result<T> = std::result::Result<T, Error>;
 // `wolfSSL_get_error`, which returns a `c_int`
 fn wolf_error_string(raw_err: std::ffi::c_ulong) -> String {
     let mut buffer = vec![0u8; wolfssl_sys::WOLFSSL_MAX_ERROR_SZ as usize];
+
+    // SAFETY:
+    // [`wolfSSL_ERR_error_string()`][0][1] is documented to store at most `WOLFSSL_MAX_ERROR_SZ` bytes, so `buffer` is appropriately sized.
```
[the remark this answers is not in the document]
expressvpn-mariappan-r: Gone with the first suggestion.

src/lib.rs
```diff
@@ -49,7 +52,10 @@ pub fn wolf_init() -> Result<()> {
 ///
 /// [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__TLS.html#function-wolfssl_cleanup
 pub fn wolf_cleanup() -> Result<()> {
-    #[allow(clippy::undocumented_unsafe_blocks)]
+    // SAFETY: [`wolfSSL_Cleanup`][0] documentation does not enforce calling this api, though it mentions this api
+    // will free memory if called.
```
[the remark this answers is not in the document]
expressvpn-mariappan-r: I have removed the SAFETY warning in `wolf_init` and `wolf_cleanup` and marked with `#[allow(clippy::undocumented_unsafe_blocks)]` for now.

Will fix up in followup PR

src/context.rs
```diff
@@ -54,6 +56,11 @@ impl ContextBuilder {
         };
 
         let result = match root {
+            // SAFETY: [`wolfSSL_CTX_load_verify_buffer`][0] ([also][1]) is documented to get the valid `ctx` pointer from wolfSSL_CTX_new().
+            // It takes buffer with size specified as `sz` and `buf` is valid, since it is from a byte slice
```
expressvpn-ian-c, line 60: This sentence doesn't quite parse correctly.

Rather than grammar tweaks however I would suggest instead something like:

> The pointer given as the `in` argument must point to a region of `sz` bytes. The values passed here are valid since they are derived from the same byte slice.

(for all instances of the sentence of course)

src/context.rs
```diff
@@ -78,6 +90,12 @@ impl ContextBuilder {
                 let path = std::ffi::CString::new(path)
                     .or(Err(Error::fatal(wolfssl_sys::WOLFSSL_BAD_PATH)))?;
                 if is_dir {
+                    // SAFETY: [`wolfSSL_CTX_load_verify_locations`][0] ([also][1]) is documented to get the valid `ctx` pointer from wolfSSL_CTX_new().
+                    // Either `file` or `path` can be valid path to a PEM file or a directory containing files.
```
expressvpn-ian-c, line 94: It seems odd to talk about "`file` or `path`" within the `is_dir` conditional (similarly in the `else` case), since we are in a context which is about a specific one of them.

Also "... can be valid path to a PEM file or a directory ..." is not quite pertinent here from a safety point of view, what's important is the validity of the pointer as a C-string -- if it happens to not actually point to a path/directory (or be an invalid path some other way) then that's not a safety issue -- i.e. WolfSSL won't crash or cause UB, it'll just return an error.

How about:

> If not NULL then the pointer passed as the `path` argument must be a valid NULL-terminated C-style string, which is guaranteed by the use of `std::ffi::CString::as_c_str()` here.

(and similar in the `else` case).

src/context.rs
```diff
@@ -111,6 +135,11 @@ impl ContextBuilder {
         let cipher_list = std::ffi::CString::new(cipher_list)
             .or(Err(Error::fatal(wolfssl_sys::WOLFSSL_FAILURE)))?;
 
+        // SAFETY: [`wolfSSL_CTX_set_cipher_list`][0] ([also][1]) is documented to get the valid `ctx` pointer from wolfSSL_CTX_new() and
+        // `list` parameter which should be a null terminated C string pointer.
```
expressvpn-ian-c, line 139: More important than what WolfSSL requires is how our code meets that requirement, i.e. it's important to mention something like "which is guaranteed by the use of `std::ffi::CString::as_c_str()` here." wording.

src/context.rs
```diff
@@ -346,6 +425,14 @@ impl Drop for Context {
     ///
     /// [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfssl_ctx_free
     fn drop(&mut self) {
+        // SAFETY: `wolfSSL_CTX_free`][0] ([also][1]) takes pointer to `WOLFSSL_CTX` and free it if the reference count becomes 0.
```
expressvpn-ian-c, line 428: Nits: missing opening `[` plus it should be "frees".

src/lib.rs
```diff
@@ -63,7 +63,11 @@ pub fn wolf_cleanup() -> Result<()> {
 #[cfg(feature = "debug")]
 pub fn enable_debugging(on: bool) {
     if on {
-        #[allow(clippy::undocumented_unsafe_blocks)]
+        // SAFETY: [`wolfSSL_Debugging_ON`][0] ([also][1]) requires `DEBUG_WOLFSSL` has to be compiled in to succeed
```
expressvpn-ian-c, line 66: nit: you don't need the `has` in the this sentence.

src/ssl.rs
```diff
@@ -196,8 +198,18 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
     /// Gets the current cipher of the session.
     /// If the cipher name is "None", return None.
     pub fn get_current_cipher_name(&self) -> Option<String> {
+        // SAFETY: [`wolfSSL_get_current_cipher`][0] ([also][1]) needs valid `wolfssl_sys::WOLFSSL` pointer as per documentation
+        // Return value is the pointer inside the ssl session. Caller can read it safely
+        //
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_get_current_cipher_suite
```
expressvpn-ian-c, line 204: linking to `wolfssl_get_current_cipher_suite` but the code is using `wolfssl_get_current_cipher`

src/ssl.rs
```diff
@@ -219,6 +231,13 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
     ///
     /// [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__TLS.html#function-wolfssl_is_init_finished
     pub fn is_init_finished(&self) -> bool {
+        // SAFETY: [`wolfSSL_CIPHER_get_name`][0] ([also][1]) expects valid pointer to `WOLFSSL` and since the `WOLFSSL` struct
+        // can be used in multiple threads based on [`Library design`][2], protected by a mutex lock
+        // Documentaion for return values seems incorrect though, having same text for both success and error case.
```
expressvpn-ian-c, line 236: typo: "Documentaion"

src/ssl.rs
```diff
@@ -351,6 +388,12 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
     pub fn try_read(&self, data_out: &mut BytesMut) -> PollResult<usize> {
         let buf = data_out.spare_capacity_mut();
 
+        // SAFETY: [`wolfSSL_read`][0] ([also][1]) expects valid pointer to `WOLFSSL` and since the `WOLFSSL` struct
+        // can be used in multiple threads based on [`Library design`][2], protected by a mutex lock
```
expressvpn-ian-c, line 392: This ought to talk about the validity of `buf` here and the corresponding length, that's the most important thing from a safety point of view I think.

src/ssl.rs
```diff
@@ -771,6 +888,8 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
             return;
         }
 
+        // SAFETY: No documentation found for `wolfSSL_dtls_set_mtu` api,
+        // From implementation, the api expects valid pointer to `WOLFSSL`
```
expressvpn-ian-c, line 892: I think you can remove the equivalent/redundant comment on the function itself.

src/ssl.rs
```diff
@@ -820,6 +952,14 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
         let domain_name = std::ffi::CString::new(domain_name.to_string())
             .expect("Input string '{domain_name:?}' contains an interior NULL");
 
+        // SAFETY: [`wolfSSL_check_domain_name`][0] ([also][1]) expects valid pointer to `WOLFSSL` and since the `WOLFSSL` struct
+        // can be used in multiple threads based on [`Library design`][2], protected by a mutex lock
+        // Documentation does not state explicitly that `dn` (domain name) should be a null terminated string
+        // Based on example (and ofcourse no `size` arg), we are constructing Cstring
```
expressvpn-ian-c, line 958: nit: missing space in "of course"

src/context.rs
```diff
@@ -35,6 +35,10 @@ impl ContextBuilder {
             .into_method_ptr()
             .ok_or(NewContextBuilderError::MethodFailed)?;
 
+        // SAFETY: [`wolfSSL_CTX_new`][0] is documented to get pointer to the valid `WOLFSSL_METHOD` structure which is created using one of the `wolfSSLvXX_XXXX_method()`.
```
expressvpn-ian-c, line 38: nit: "...to a valid..."

src/context.rs
```diff
@@ -60,6 +69,11 @@ impl ContextBuilder {
                     WOLFSSL_FILETYPE_ASN1,
                 )
             },
+            // SAFETY: [`wolfSSL_CTX_load_verify_buffer`][0] ([also][1]) is documented to get the valid `ctx` pointer from wolfSSL_CTX_new().
```
expressvpn-ian-c, line 72: it should be "a valid" again (not "the"), although I'd perhaps word this as "[`wolfSSL_CTX_load_verify_buffer`][0] ([also][1]) requires a valid `ctx` pointer from `wolfSSL_CTX_new()`.

src/ssl.rs
```diff
@@ -156,6 +156,10 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
         ctx: &Context,
         config: SessionConfig<IOCB>,
     ) -> std::result::Result<Self, NewSessionError> {
+        // SAFETY: [`wolfSSL_new`][0] ([also][1]) needs valid `wolfssl_sys::WOLFSSL_CTX` pointer as per documentation
```
expressvpn-ian-c, line 159: "a valid".

src/ssl.rs
```diff
@@ -233,6 +254,12 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
     ///
     /// [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_negotiate
     pub fn try_negotiate(&self) -> PollResult<()> {
+        // SAFETY: [`wolfSSL_negotiate`][0] ([also][1]) expects valid pointer to `WOLFSSL` and since the `WOLFSSL` struct
+        // can be used in multiple threads based on [`Library design`][2], protected by a mutex lock
```
expressvpn-ian-c, line 258: The "and since  ... protected by a mutex lock" clause is pretty hard to grok IMHO.

How about:

> ... expects a valid pointer to a `WOLFSSL`. Per the [`Library design`][2] access is synchronized via the containing [`Mutex`]?

expressvpn-ian-c (commented):
Sorry, some more comments, none major I don't think (although some are quite repetitive because the same text appears a number of times).

I went back and forth a few times on how I felt about:

> ... is documented to get the valid `ctx` pointer from wolfSSL_CTX_new().

and

> ... expects valid pointer to `WOLFSSL` and since the `WOLFSSL` struct can be used in multiple threads based on [`Library design`][2], protected by a mutex lock

which explain the C API requirements, which in turn is the requirement for Safety, but how the Rust code meets those requirements is not mentioned.

If we were being super precise we'd say something like "the pointer is always valid by construction of the Foo type". OTOH maybe that's so obviously a property of the use of the `as_ptr()` method that it's not worth saying. I eventually landed on the not worth it side, mainly because the [stdlib guidance](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) says:

> Inside safe elements, a SAFETY: comment must not depend on anything from the caller beside properly constructed types and values (i.e, if your function receives a reference that is unaligned or null, it is the caller fault if it fails and not yours).

So we can take it as written that `self.ctx` is a valid `NonNull<wolfssl_sys::WOLFSSL_CTX>` and that it isn't `NULL`.

src/ssl.rs
expressvpn-mariappan-r: Updated in all occurrences of the above statement.

src/ssl.rs
```diff
@@ -486,6 +563,10 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
             e => unreachable!("{e:?}"),
         }
 
+        // SAFETY: Based on `wolfSSL_key_update_response`][0], required will be populated if the api returns success.
+        // So safety to call `assume_init()` on success case. On error case we panic!
+        //
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_key_update_response]
```
expressvpn-ian-c, line 569: Nit: trailing `]`.

src/ssl.rs
```diff
@@ -486,6 +563,10 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
             e => unreachable!("{e:?}"),
         }
 
+        // SAFETY: Based on `wolfSSL_key_update_response`][0], required will be populated if the api returns success.
+        // So safety to call `assume_init()` on success case. On error case we panic!
```
expressvpn-ian-c, line 567: I got confused here by "On error case we panic", thinking this meant in the match below -- but rereading I think you were talking about the previous match. Can we say "On error we already panicked above" to clarify?

(Aside: I think the use of `e` for the blanket match below is misleading, and contributed to my misreading here, that use of `e` is not "an error" it's "a value other than 0 or 1", IMHO `nr` would be a better name).

expressvpn-ian-c (commented):
Sorry, I thought I clicked the submit button yesterday but clearly forgot.

src/ssl.rs
expressvpn-mariappan-r: Yep updated the comment!
But the `e` on blanket match of `assume_init` looks ok to me.
Based on documentation`wolfSSL_key_update_response` api which is supposed to initialize the `required` variable guarantees the value will only be `0` or `1`

expressvpn-ian-c (approved):

PR #27, opened 2023-09-12
Use specific errors when creating a `ContextBuilder` or a `Session` fails

expressvpn-ian-c:
This allows a bit better error reporting than returning an `Option`. I think mainly useful in practice for `Session` in the case where one of the post-creation setup methods fails, WolfSSL doesn't give us any hint why the actual creates fail.

### DIFF
--- a/src/ssl.rs
+++ b/src/ssl.rs
@@ -941,7 +941,7 @@ mod tests {
         server_protocol: Protocol,
     ) -> (TestClient, TestClient) {
         let client_ctx = ContextBuilder::new(client_protocol)
-            .unwrap_or_else(|| panic!("new({client_protocol:?})"))
+            .unwrap_or_else(|e| panic!("new({client_protocol:?}): {e}"))
             .with_root_certificate(RootCertificate::Asn1Buffer(CA_CERT))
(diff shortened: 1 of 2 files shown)


expressvpn-pete-m (dismissed):

expressvpn-mariappan-r (dismissed):
LGTM

src/ssl.rs
```diff
@@ -919,15 +941,15 @@ mod tests {
...
+            .unwrap_or_else(|e| panic!("new({client_protocol:?}): {e}"))
             .with_root_certificate(RootCertificate::Asn1Buffer(CA_CERT))
             .unwrap()
             .with_secure_renegotiation()
             .unwrap()
             .build();
 
         let server_ctx = ContextBuilder::new(server_protocol)
-            .unwrap_or_else(|| panic!("new({server_protocol:?})"))
+            .unwrap_or_else(|e| panic!("new({client_protocol:?}): {e}"))
```
expressvpn-mariappan-r, line 952:
```suggestion
            .unwrap_or_else(|e| panic!("new({server_protocol:?}): {e}"))
```

PR #30, opened 2023-09-14
Remove `wolf_init` from public view and remove `wolf_cleanup` entirely

expressvpn-ian-c:
As discussed in https://github.com/expressvpn/wolfssl/pull/24#discussion_r1322705812 https://github.com/expressvpn/wolfssl/pull/24#discussion_r1322685417 these put safety were putting safety requirements into the responsibility of the caller.

For `wolf_init` push the responsibility for calling this into this crate so we can be sure we meet the "before any other call to the library" requirement from [the documentation](https://www.wolfssl.com/doxygen/group__TLS.html#ga789ef74e34df659a62f06da2ea709737). The places where we need to do the initialization are few in number and none are in hot paths -- I think it unlikely that there would ever be a hot path which didn't require some cold path setup first so I think we aren't making a rod for our own back by taking on this responsibility, although we will have to be careful to remember to add calls to `wolf_init` as we add new top-level entry points.

We then use a [`std::sync::OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) to meet the "must be called once per application" requirement (this might be too strict, the docs don't say "at most once", but note that a second unsuccessful call to `wolfSSL_Init` would undo a prior success though!).

Since we have no use case for `wolf_cleanup` currently simply remove it so we don't have to reason about when/how it would be ok to call.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -451,7 +451,7 @@ impl Drop for Context {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{wolf_cleanup, wolf_init};
+    use crate::wolf_init;
     use test_case::test_case;
 
     #[test_case(Protocol::DtlsClient)]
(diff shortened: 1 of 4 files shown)


expressvpn-pete-m (approved):

expressvpn-mariappan-r (approved):
LGTM

PR #37, opened 2023-09-22
Do not construct `Error` values unless needed.

xv-ian-c:
`Error::fatal` does a fair bit of work, including calling `wolf_error_string` to stringify the error (which allocates), lets not do that until an error has actually happened.

Noticed when setting a breakpoint on `Error::fatal` which then triggered frequently.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -94,9 +94,9 @@ impl ContextBuilder {
                 let is_dir = path.is_dir();
                 let path = path
                     .to_str()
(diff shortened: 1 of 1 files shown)


kp-mariappan-ramasamy (approved):
Nice find! LGTM

xv-ian-c:
Thanks, I'm a bit surprised clippy didn't suggest this -- I wonder if the fact that it is a method on an enum makes clippy thinks it must be cheap.

PR #39, opened 2023-09-25
Convert to cargo workspace and integrate wolfssl-sys

xv-ian-c:
Some of the adjustments were taken from https://github.com/expressvpn/wolfssl-sys/pull/53.

PR #42, opened 2023-09-28
Enable DTLS1.3

kp-mariappan-ramasamy:
Enable DTLS1.3 while building wolfssl-sys and create Protocol variants for it

### DIFF
--- a/wolfssl-sys/build.rs
+++ b/wolfssl-sys/build.rs
@@ -53,6 +53,8 @@ fn build_wolfssl(dest: &str) -> PathBuf {
         .disable_shared()
         // Enable TLS/1.3
         .enable("tls13", None)
+        // Enable DTLS/1.3
+        .enable("dtls13", None)
         // Disable old TLS versions
         .disable("oldtls", None)
         // Enable single threaded mode
--- a/wolfssl/src/lib.rs
+++ b/wolfssl/src/lib.rs
@@ -93,10 +93,14 @@ pub enum Protocol {
     DtlsClient,
     /// `wolfDTLSv1_2_client_method`
     DtlsClientV1_2,
+    /// `wolfDTLSv1_3_client_method`
+    DtlsClientV1_3,
     /// `wolfDTLS_server_method`
     DtlsServer,
     /// `wolfDTLSv1_2_server_method`
     DtlsServerV1_2,
+    /// `wolfDTLSv1_3_server_method`
+    DtlsServerV1_3,
     /// `wolfTLS_client_method`
     TlsClient,
     /// `wolfTLSv1_2_client_method`
@@ -127,9 +131,13 @@ impl Protocol {
             // SAFETY: as above
             Self::DtlsClientV1_2 => unsafe { wolfssl_sys::wolfDTLSv1_2_client_method() },
             // SAFETY: as above
+            Self::DtlsClientV1_3 => unsafe { wolfssl_sys::wolfDTLSv1_3_client_method() },
+            // SAFETY: as above
             Self::DtlsServer => unsafe { wolfssl_sys::wolfDTLS_server_method() },
             // SAFETY: as above
             Self::DtlsServerV1_2 => unsafe { wolfssl_sys::wolfDTLSv1_2_server_method() },
+            // SAFETY: as above
+            Self::DtlsServerV1_3 => unsafe { wolfssl_sys::wolfDTLSv1_3_server_method() },
             // SAFETY: as above
             Self::TlsClient => unsafe { wolfssl_sys::wolfTLS_client_method() },
             // SAFETY: as above


wolfssl/tests/ping_pong.rs
```diff
@@ -192,18 +193,21 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
     println!("[Server] Finished");
 }
 
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2)]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_3)]
+// TODO: DTLS1.3 <-> DTLS1.2 is not working yet
```
xv-ian-c, line 198: Do we have a link to an upstream bug?

My (second hand) understanding was the client 1.2 <-> server 1.3 was ok and only client 1.3 <-> server 1.2 had issues?

wolfssl/tests/ping_pong.rs
```diff
@@ -192,18 +193,21 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
     println!("[Server] Finished");
 }
 
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2)]
```
xv-ian-c, line 196: `test_case` will derive the test name from this here you get `dtls::protocol_dtlsclientv1_3_protocol_dtlsserverv1_3_expects`.

You can specificy a particular name with

```
[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2; "client_1.2_server_1.3")]
```
if we want to make it a little more readable.

xv-ian-c (commented):
LGTM, would just like to clarify the nature of the compatibility issue.

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: Oh nice! will update it

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: No, I don't have an exact link to it. A quick search did not find any.
I tested both and it was panicking.
Wanted to revisit this, after `lightway-core` got fixed first, so added this TODO

kp-mariappan-ramasamy: To understand the behavior further, I have extended the tests for TLS too which fails, if we mix protocols:

```
---- tls::client_1_2_server_1_3 stdout ----
[Client] Connecting...
[Server] Connecting...
thread 'tls::client_1_2_server_1_3' panicked at '[Server] try_negotiate: Fatal(FatalError { what: "record layer version error", code: -326 })', wolfssl/tests/ping_pong.rs:169:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tls::client_1_3_server_1_2 stdout ----
[Client] Connecting...
[Server] Connecting...
thread 'tls::client_1_3_server_1_2' panicked at '[Client] try_negotiate: Fatal(FatalError { what: "malformed buffer input error", code: -328 })', wolfssl/tests/ping_pong.rs:121:47
```

kp-mariappan-ramasamy: I found the issue, `Protocol::::DtlsServerV1_3` aka. `wolfssl_sys::wolfDTLSv1_3_server_method()` does not support backward compatibility, we need to use `Protocol::DtlsServer` to support it.

Updated the PR with testcases for all the combinations.

For these cases, panics are expected since there is no backward compatability
`Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_2 => panics;`
`Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_3 => panics;`


Still DTLS client 1.2/1.3 <-> server 1.2 has some issues similar to lightway-core.

xv-ian-c: Well spotted, I had misunderstood what the versioned methods were about, it makes sense as you describe it.

wolfssl/tests/ping_pong.rs
```diff
@@ -192,18 +193,27 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
...
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_3 => panics; "client_1.2_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2; "client_1.2_server_1.2")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServer; "client_1.2_server")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_3; "client_1.3_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_2 => panics; "client_1.3_server_1.2")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServer; "client_1.3_server")]
+#[test_case(Protocol::DtlsClient, Protocol::DtlsServerV1_3; "client_server_1.3")]
+// TODO: WolfSSL downgrade bug
+// #[test_case(Protocol::DtlsClient, Protocol::DtlsServerV1_2; "client_server_1.2")]
+#[test_case(Protocol::DtlsClient, Protocol::DtlsServer; "client_server")]
```
xv-ian-c: nit: You don't need a blank line between the cases and the tokio tag.

wolfssl/tests/ping_pong.rs
```diff
@@ -192,18 +193,27 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
     println!("[Server] Finished");
 }
 
-#[tokio::test]
-async fn dtls() {
-    use Protocol::*;
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_3 => panics; "client_1.2_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2; "client_1.2_server_1.2")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServer; "client_1.2_server")]
```
xv-ian-c: WDYT of including `any` in place of the version here to clarify and make the pattern consistently `(client|server)_($version|any)`?

wolfssl-sys/build.rs
```diff
@@ -40,6 +40,37 @@ fn copy_wolfssl(dest: &str) -> std::io::Result<()> {
...
+    let patch_buffer = Command::new("cat")
+        .arg(patch)
+        .stdout(Stdio::piped())
+        .spawn()
+        .unwrap();
+    Command::new("patch")
+        .arg("-d")
+        .arg(wolfssl_path)
+        .arg("-p1")
+        .stdin(Stdio::from(patch_buffer.stdout.unwrap()))
```
xv-ian-c, line 69: I think you can drop the `.stdin()` and the needless use of `cat` and instead name the file on the command line i.e. just `.arg(patch)`.

If for some reason the patch really has to be given on stdin please use [`std::fs::File::open`](https://doc.rust-lang.org/std/fs/struct.File.html#method.open) not `cat`.

wolfssl-sys/build.rs
```diff
@@ -40,6 +40,37 @@ fn copy_wolfssl(dest: &str) -> std::io::Result<()> {
...
+    "patches/1001-Allow-easily-disabling-of-SPHINCS.patch",
+    "patches/2000-Rename-utils.c-to-utils.h.patch",
+    "patches/2001-Merge-pull-request-6700-from-julek-wolfssl-dtls13-do.patch",
+];
+
+/**
+ * Apply patch to wolfssl-src
+ */
+fn apply_patch(dest: &str, patch: &str) {
+    let wolfssl_path = format!("{dest}/wolfssl-src");
```
xv-ian-c, line 58: Strictly speaking I think you should emit a [`[cargo:rerun-if-changed=PATH](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath)`)https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) for each patch file.

But we aren't retriggering if the wolfssl-src code itself changes (and I think that would be hard to achieve since it is a _lot_ of files) so maybe we can just gloss over this.

This is slightly compounding the kinds of issues I was worried about in https://github.com/expressvpn/wolfssl-sys/pull/53#discussion_r1332631073 (which I carried over into #39 here), one day maybe we'll want to revisit this but for now I think we can cope.

wolfssl/tests/ping_pong.rs
```diff
@@ -219,18 +219,25 @@ async fn dtls(client_protocol: Protocol, server_protocol: Protocol) {
...
+#[test_case(Protocol::TlsClientV1_3, Protocol::TlsServerV1_2 => panics; "client_1.3_server_1.2")]
+#[test_case(Protocol::TlsClientV1_3, Protocol::TlsServer; "client_1.3_server")]
+#[test_case(Protocol::TlsClient, Protocol::TlsServerV1_3; "client_server_1.3")]
+#[test_case(Protocol::TlsClient, Protocol::TlsServerV1_2; "client_server_1.2")]
+#[test_case(Protocol::TlsClient, Protocol::TlsServer; "client_server")]
 #[tokio::test]
-async fn tls() {
-    use Protocol::*;
-
+async fn tls(client_protocol: Protocol, server_protocol: Protocol) {
```
xv-ian-c: Nit: commit message should say "extend" not "extent"

wolfssl/tests/ping_pong.rs
```diff
@@ -192,18 +193,27 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
     println!("[Server] Finished");
 }
 
-#[tokio::test]
-async fn dtls() {
-    use Protocol::*;
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_3 => panics; "client_1.2_server_1.3")]
```
xv-ian-c: The `panics` matcher takes an optional string and asserts that it appears in the panic message.

I've been tripped up in the past by code which panicked for some other reason than the one we were expecting, so adding `"record layer version error"` (or whatever unique-ish string there is) might be wise for these cases (and the TLS ones)

wolfssl/src/ssl.rs
```diff
@@ -714,6 +714,29 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+        // SAFETY: [`wolfSSL_dtls13_use_quick_timeout`][0] ([also][1]) expects a valid pointer to `WOLFSSL`. Per the
+        // [Library design][2] access is synchronized via the containing [`Mutex`]
+        //
+        // [0]: https://www.wolfssl.com/documentation/manuals/wolfssl/ssl_8h.html#function-wolfssl_dtls13_use_quick_timeout
+        // [1]: https://www.wolfssl.com/doxygen/ssl_8h.html#a61f3b53cb0397dd1debc8b8daaa490c2
+        // [2]: https://www.wolfssl.com/documentation/manuals/wolfssl/chapter09.html#thread-safety
+        match unsafe { wolfssl_sys::wolfSSL_dtls13_use_quick_timeout(ssl.as_ptr()) } {
+            0 => false,
+            1 => true,
+            e => unreachable!("{e:?}"),
```
xv-ian-c, line 759: Docs say:

> Return: true if the application code should setup a quicker timeout

In C true is any non-zero value not specifically 1. So I think:
```rust
match ... {
    0 => false
    _ => true
}
```

might be closer to what the API docs say?

Or equivalently a `!= 0` check. Bit of a shame there is no `From<c_int> for bool`, but it is what it is.

wolfssl/tests/ping_pong.rs
```diff
@@ -193,51 +203,58 @@ async fn server<S: SockIO>(sock: S, protocol: Protocol) {
...
-#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_2 => panics; "client_1.3_server_1.2")]
-#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServer; "client_1.3_server")]
-#[test_case(Protocol::DtlsClient, Protocol::DtlsServerV1_3; "client_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_3, ProtocolVersion::Unknown => panics; "client_1.2_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServerV1_2, ProtocolVersion::DtlsV1_2; "client_1.2_server_1.2")]
+#[test_case(Protocol::DtlsClientV1_2, Protocol::DtlsServer, ProtocolVersion::DtlsV1_2; "client_1.2_server")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_3, ProtocolVersion::DtlsV1_3; "client_1.3_server_1.3")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServerV1_2, ProtocolVersion::Unknown => panics; "client_1.3_server_1.2")]
+#[test_case(Protocol::DtlsClientV1_3, Protocol::DtlsServer, ProtocolVersion::DtlsV1_3; "client_1.3_server")]
+#[test_case(Protocol::DtlsClient, Protocol::DtlsServerV1_3, ProtocolVersion::DtlsV1_3; "client_server_1.3")]
```
xv-ian-c, line 212: Shall we consider adding whitespace after each `,` and the `;` to make these more tabular (=> readable)?

xv-ian-c (commented):
Nice!

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: Yep, I know. I added it for readabiliy.
But cargo fmt does not like that and removed it anyway. So it s not there in the final commit.

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: Yep, sounds good

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: Nice suggestion. Will incorporate it

wolfssl/src/ssl.rs
kp-mariappan-ramasamy: All the existing code follows this convention of converting <c_int> to bool. 
Think its better to be consistent. We can followup with a Jira/PR, if we want to update all.

xv-ian-c: Other cases e.g. https://www.wolfssl.com/documentation/manuals/wolfssl/group__TLS.html#function-wolfssl_is_init_finished or https://www.wolfssl.com/documentation/manuals/wolfssl/ssl_8h.html#function-wolfssl_dtls the docs do specifically say 0 and 1. This one doesn't.

wolfssl-sys/build.rs
kp-mariappan-ramasamy: Yep, you are right about wolfssl source.

If we think logically, this may not be needed for patches.Whenever we add a new patch, we will update the build script and it will rebuild. That's why we have a static list of patches in script instead of reading the directory and apply all patches.
And we will almost never hand edit a patch. Either add it or remove it.

But anyway I have added for completeness.

wolfssl-sys/build.rs
kp-mariappan-ramasamy: I see, you are not a :cat:  person
Got rid of it now

wolfssl-sys/build.rs
xv-ian-c: (One of) the problem(s) is reapplying the same patch a second time will fail. But anyway, the status quo is already fragile IMHO and the changes here don't change that so :shrug:

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: I do not want to go down this route of hand editing the source code as tabular format.
IMHO it might be useful for readme and documentation, but not for test cases.
I prefer to follow what `cargo fmt` provides

xv-ian-c: `cargo fmt` doesn't touch the contents of `test_case(...)` so you won't be fighting it here.

IME representing test cases as tables is a really helpful technique and it works nicely in the source e.g. https://github.com/xvpn/password_manager/blob/master/pmcore/src/tests/unlock.rs#L529

wolfssl/src/ssl.rs
kp-mariappan-ramasamy: I think the other way. 

In C, true is actually `1` if we use stdbool. Using any value other than 0 for true is a convention or hack.

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdbool.h.html
https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/arm/arm-eabi-4.8/+/4cedbb07c5475ce9a5fc9f2d9645c00292e4be72/lib/gcc/arm-eabi/4.8/include/stdbool.h#34

If we go by this logic, think the changes are relevant. 
Also it is consistent :man_shrugging:

xv-ian-c: There is no stdbool involved here:
```
WOLFSSL_API int wolfSSL_dtls13_use_quick_timeout(WOLFSSL* ssl);
```

The treatment of an integer type a boolean context in C, where "non-zero is true", is neither a hack nor merely a convention, it's part of the definition of the language.

WRT consistency, the wolfssl API is inconsistent here it seems, so it follows that we will be too.

wolfssl/tests/ping_pong.rs
kp-mariappan-ramasamy: IMHO its not worth the time to manually beautify the `test` code unless a tool does it.
For example:
https://github.com/xvpn/lightway/blob/main/lightway-core/src/wire.rs#L491

We have lot of test cases here, which will be clumsy. We can spend of time making it readable.
But it may not be worth it.

xv-ian-c: IMHO it is, in general, absolutely worth it, a tabular form makes it much easier to see if all cases are covered -- I made this comment because I was doing exactly that.

There are exceptions and I think the link you gave is to one of them, aliging things there would result in unmanageable line lengths.

wolfssl/src/ssl.rs
kp-mariappan-ramasamy: Ack! Most C library function return true(nonzero) value in general
So updated the code as per above suggestion

xv-ian-c (approved):

PR #45, opened 2023-09-29
LIT-122: Add support for `earthly doc`

xv-andre-l:
Here's what it looks like!

```
$ earthly doc
TARGETS:
  +build-dev
      build-dev builds with the Cargo dev profile and produces debug artifacts
  +build-release
      build-release builds with the Cargo release profile and produces release artifacts
  +run-tests
      run-tests executes all unit and integration tests via Cargo
  +run-coverage
      run-coverage generates a report of code coverage by unit and integration tests
  +build
      build runs tests and then creates a release build
  +lint
      lint runs cargo clippy on the ssource code
  +fmt
      fmt formats Rust code according to style guidelines
  +check-dependencies
      check-dependencies lints our dependencies
```

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -18,20 +18,24 @@ copy-src:
     FROM +build-deps
     COPY --dir Cargo.toml Cargo.lock deny.toml wolfssl wolfssl-sys ./
 
+# build-dev builds with the Cargo dev profile and produces debug artifacts
 build-dev:
     FROM +copy-src
     DO rust-udc+CARGO --args="build"
     SAVE ARTIFACT target/debug /debug AS LOCAL artifacts/debug
 
+# build-release builds with the Cargo release profile and produces release artifacts
 build-release:
     FROM +copy-src
     DO rust-udc+CARGO --args="build --release"
     SAVE ARTIFACT target/release /release AS LOCAL artifacts/release
 
+# run-tests executes all unit and integration tests via Cargo
 run-tests:
     FROM +copy-src
     DO rust-udc+CARGO --args="test"
 
+# run-coverage generates a report of code coverage by unit and integration tests via cargo-llvm-cov
 run-coverage:
     FROM +copy-src
     DO rust-udc+CARGO --args="llvm-cov test"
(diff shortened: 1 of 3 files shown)


Earthfile
```diff
@@ -42,23 +46,28 @@ run-coverage:
     DO rust-udc+CARGO --args="llvm-cov report --html --output-dir /tmp/coverage/"
     SAVE ARTIFACT /tmp/coverage/*
 
+# build runs tests and then creates a release build
 build:
     BUILD +run-tests
     BUILD +build-release
 
+# build-create creates a .crate file for distribution of source code
```
xv-ian-c, line 54: Typo: `create` instead of `crate` (I do this 90% of the time I try to write crate...).

Given the display context:
```
  +build
      build runs tests and then creates a release build
```

We should consider eliding the target name, e.g.

```
  +build
      runs tests and then creates a release build
```

I think that works in the source too:
```
# runs tests and then creates a release build
build:
```

The rule would be "the doc comment should be a sentence fragment suitable to be appended to the target name".

Earthfile
```diff
@@ -42,23 +46,28 @@ run-coverage:
...
     FROM +copy-src
     DO rust-udc+CARGO --args="package"
     SAVE ARTIFACT target/package/*.crate /package/ AS LOCAL artifacts/crate/
 
+# lint runs cargo clippy on the source code
 lint:
     FROM +copy-src
     DO rust-udc+CARGO --args="clippy --all-features --all-targets -- -D warnings"
 
+# fmt formats Rust code according to style guidelines
```
kp-mariappan-ramasamy, line 65:
```suggestion
# fmt checks whether Rust code is formatted according to style guidelines
```
This actually does not format the code. It only checks whether format is good.

Earthfile
```diff
@@ -18,20 +18,24 @@ copy-src:
...
     DO rust-udc+CARGO --args="build"
     SAVE ARTIFACT target/debug /debug AS LOCAL artifacts/debug
 
+# build-release builds with the Cargo release profile and produces release artifacts
 build-release:
     FROM +copy-src
     DO rust-udc+CARGO --args="build --release"
     SAVE ARTIFACT target/release /release AS LOCAL artifacts/release
 
+# run-tests executes all unit and integration tests via Cargo
```
kp-mariappan-ramasamy, line 33: `via Cargo` may not be needed since all the targets are via cargo

expressvpn-andre-l: I think `via Cargo` is useful because we can make this information available to any user of `earthly doc` without needing the user dig into the `Earthfile` itself. 

I did wonder whether it was too much detail to specify `via Cargo`, but I kept it because:
1) https://docs.earthly.dev/docs/earthly-command#earthly-doc examples also have the same amount of detail about `go build`
2) Comments that are out of sync with the implementation are confusing, but I doubt the comment will easily get out of sync with the implementation.  That's because the `DO rust-udc+CARGO --args="test"` is a few lines after.  If we change the implementation to now use Cargo anymore, I'm sure it will be obvious to update that comment.
3) I see from https://doc.rust-lang.org/cargo/commands/cargo-test.html that Cargo picks a bunch of defaults for you when running tests, so I think letting the user know that it's `via Cargo` is important, because that clues them in that the behaviour of the command will be according to Cargo's defaults.

Earthfile
expressvpn-andre-l: I've fixed the `Typo: create instead of crate` with another push :)

Earthfile
expressvpn-andre-l: Thanks, fixed!

Earthfile
kp-mariappan-ramasamy: Yeah sounds good. 
Can we please add about `cargo-llvm` and `cargo deny` in their respective targets to be uniform ?

expressvpn-andre-l: Sure thing, updated the PR!

Earthfile
xv-ian-c: If there is a style guide which we can follow then we should do so!

xv-ian-c (approved):
LGTM.

If I were nitpicking I'd say that "cargo-foo" is normally invoked/spelled as "cargo foo" and that the Earthfile example seems to put single quotes around literal commands which are run when describing them. Not a big deal though IMHO.

PR #52, opened 2023-10-06
upgrade earthly and cargo UDC

idelvall:
This PR:
- Upgrades the version of earthly to 0.7.20
- Uses the a new UDC version that removes the usage of CACHEs in favor of mount caches, for better performance when the cache size grows.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: earthly/actions-setup@v1
         with:
-          version: v0.7.19
+          version: v0.7.20
       - uses: actions/checkout@v3
         with:
           submodules: true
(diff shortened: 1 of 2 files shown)


wolfssl/src/context.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
 use std::ptr::NonNull;
 use thiserror::Error;
 
-/// Produces a [`Context`] once built.
+/// Produces  a [`Context`] once built.
```
xv-ian-c, line 10: Can we drop this commit please.

xv-ian-c (commented):
CI changes are good, but please drop the spurious change.

wolfssl/src/context.rs
idelvall: done!

xv-ian-c (approved):

PR #58, opened 2023-10-16
Update to new Earthly UDC

xv-ian-c:
- Bumping to https://github.com/earthly/lib/compare/2.2.4...2.2.5 allow us to use `--keep_fingerprints=true` since after 45423285950 this workaround is not needed.
- Bumping to https://github.com/earthly/lib/compare/2.2.5...2.2.6 keeps us up to date, but since we do not use `--output` doesn't affect us here.
- Use the UDC to install `cargo-deny` and `cargo-llvm-cov` so we will cache things
- Use a matrix job for the simple "just run Earthly target" cases, less repetition

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,13 +6,17 @@ on:
 
 jobs:
   ci:
-    needs: [test, build, fmt, lint, coverage, check-dependencies]
+    needs: [earthly, coverage]
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


kp-mariappan-ramasamy (dismissed):

xv-ian-c:
We need some more UDC updates, to draft in the meantime.

kp-mariappan-ramasamy (approved):

PR #61, opened 2023-10-24
wolfssl: Allow key rotation for DTLS 1.3 connections

xv-ian-c:
Although the [documentation] talks about "TLS 1.3" this feature is available for DTLS 1.3 too (confirmed in the code).

[documentation]: https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_update_keys

### DIFF
--- a/wolfssl/src/lib.rs
+++ b/wolfssl/src/lib.rs
@@ -202,6 +202,11 @@ impl Protocol {
     fn is_tls_13(&self) -> bool {
         matches!(self, Self::TlsClientV1_3 | Self::TlsServerV1_3)
     }
+
+    /// Checks if the protocol is compatible with DTLS 1.3
+    fn is_dtls_13(&self) -> bool {
+        matches!(self, Self::DtlsClientV1_3 | Self::DtlsServerV1_3)
+    }
 }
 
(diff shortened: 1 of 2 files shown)


wolfssl/src/lib.rs
```diff
@@ -202,6 +202,11 @@ impl Protocol {
     fn is_tls_13(&self) -> bool {
         matches!(self, Self::TlsClientV1_3 | Self::TlsServerV1_3)
     }
+
+    /// Checks if the method is compatible with DTLS 1.3
```
kp-mariappan-ramasamy, line 206: method -> Protocol

kp-mariappan-ramasamy (approved):
LGTM 
Minor nit

wolfssl/src/lib.rs
xv-ian-c: It's an odd one this, the WolfSSL naming is `wolfDTLS_*_method` but the type here is `Protocol` (which is a better name for our API IMHO).

I think you are right though and the same thing applies to the existing `is_tls_13` method.

kp-mariappan-ramasamy (approved):

PR #70, opened 2023-11-07
LIT-193: New trait for receiving TLS1.3 secrets in debug mode

kp-mariappan-ramasamy:
WolfSSL provides  a way to dump secrets in TLS13 which can be used to decrypt traffic using Wireshark.

Ref: 
https://github.com/wolfSSL/wolfssl-examples/blob/master/tls/server-tls13.c 
https://github.com/wolfSSL/wolfssl-examples/blob/master/tls/client-tls13.c 

This PR adds helpers in wolfssl-rs to support that feature

### DIFF
--- a/wolfssl/src/debug.rs
+++ b/wolfssl/src/debug.rs
@@ -1,0 +1,105 @@
+#![cfg(feature = "debug")]
+
+use std::os::raw::{c_int, c_uint};
+use std::sync::Arc;
+
+/// Application provided callbacks to receive TLS1.3 secrets
+///
+/// There are two apis in this trait.
+/// Application can opt to receive either raw secrets or String which is suitable to use in
+/// WireShark keylog
+/// <https://www.wireshark.org/docs/wsug_html_chunked/ChIOExportSection.html#ChIOExportTLSSessionKeys>
+/// <https://wiki.wireshark.org/TLS#using-the-pre-master-secret>
+pub trait Tls13SecretCallbacks {
+    /// Called when WolfSSL wishes to send new TLS1.3 secret used
+    /// `secret` is formatted as WireShark keylog string.
+    /// It can be saved to a file and used in WireShark directly
+    fn wireshark_keylog(&self, _secret: String);
+
+    /// Called when WolfSSL wishes to send new TLS1.3 secret used
+    /// Random value and secrets along with secret type is send separately
+    /// The default implementation parses the secret/random and generate the
+    /// WireShark compatible keylog string and call `Tls13SecretCallbacks::wireshark_keylog`.
+    ///
+    /// Application can choose to override this to get the secret/random directly.
+    fn secrets(&self, secret_type: Tls13Secret, random: &[u8], secret: &[u8]) {
+        let mut keylog = secret_type.to_string();
+        keylog.push(' ');
+
+        random
+            .iter()
+            .for_each(|i| keylog.push_str(&format!("{:02x}", i)));
+        keylog.push(' ');
+
+        secret
+            .iter()
+            .for_each(|f| keylog.push_str(&format!("{:02x}", f)));
+        keylog.push('\n');
+
+        self.wireshark_keylog(keylog);
+    }
+}
+
+/// Convenience type to use as function arguments
+pub type Tls13SecretCallbacksArg = Arc<dyn Tls13SecretCallbacks + Send + Sync>;
+
+pub(crate) const RANDOM_SIZE: usize = 32;
+
+/// Tls13 Secret types
+/// To be used in Wireshark
+pub enum Tls13Secret {
+    /// "CLIENT_EARLY_TRAFFIC_SECRET"
+    ClientEarlyTrafficSecret,
+    /// "CLIENT_HANDSHAKE_TRAFFIC_SECRET"
+    ClientHandshakeTrafficSecret,
+    /// "SERVER_HANDSHAKE_TRAFFIC_SECRET"
+    ServerHandshakeTrafficSecret,
+    /// "CLIENT_TRAFFIC_SECRET_0"
+    ClientTrafficSecret,
+    ///"SERVER_TRAFFIC_SECRET_0"
+    ServerTrafficSecret,
+    /// "EARLY_EXPORTER_SECRET"
+    EarlyExporterSecret,
+    ///
+    ExporterSecret,
+    /// "EXPORTER_SECRET"
+    UnknownSecret(c_uint),
+}
+
+impl ToString for Tls13Secret {
+    fn to_string(&self) -> String {
+        use Tls13Secret::*;
+        match self {
+            ClientEarlyTrafficSecret => "CLIENT_EARLY_TRAFFIC_SECRET",
+            ClientHandshakeTrafficSecret => "CLIENT_HANDSHAKE_TRAFFIC_SECRET",
+            ServerHandshakeTrafficSecret => "SERVER_HANDSHAKE_TRAFFIC_SECRET",
+            ClientTrafficSecret => "CLIENT_TRAFFIC_SECRET_0",
+            ServerTrafficSecret => "SERVER_TRAFFIC_SECRET_0",
+            EarlyExporterSecret => "EARLY_EXPORTER_SECRET",
+            ExporterSecret => "EXPORTER_SECRET",
+            UnknownSecret(_e) => "UNKNOWN_SECRET",
+        }
+        .to_string()
+    }
+}
+
+impl From<c_int> for Tls13Secret {
+    fn from(value: c_int) -> Self {
+        match value as c_uint {
+            wolfssl_sys::Tls13Secret_CLIENT_EARLY_TRAFFIC_SECRET => {
+                Tls13Secret::ClientEarlyTrafficSecret
+            }
+            wolfssl_sys::Tls13Secret_CLIENT_HANDSHAKE_TRAFFIC_SECRET => {
+                Tls13Secret::ClientHandshakeTrafficSecret
+            }
+            wolfssl_sys::Tls13Secret_SERVER_HANDSHAKE_TRAFFIC_SECRET => {
+                Tls13Secret::ServerHandshakeTrafficSecret
+            }
+            wolfssl_sys::Tls13Secret_CLIENT_TRAFFIC_SECRET => Tls13Secret::ClientTrafficSecret,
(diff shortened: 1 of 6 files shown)


wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1111,95 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+
+    /// Enable key logging for applications
+    #[cfg(feature = "debug")]
+    pub fn enable_key_logging(&mut self, keylog: Arc<dyn Tls13SecretCallbacks>) -> Result<()> {
+        self.keylog = Some(keylog.clone());
+
+        let keylog_c = Box::into_raw(Box::new(keylog));
+
+        // SAFETY: No documentation available for [`wolfSSL_KeepArrays`] [`wolfSSL_set_tls13_secret_cb`].
+        // But based on api implementation, it expects a valid pointer to `WOLFSSL`.
```
xv-ian-c, line 1123: I think we could use some discussion of the safety aspects relating to the lifetime of `keylog_c` and the memory behind the keylog `Arc`.

Do you need the box here, I think [`Arc::as_ptr`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.as_ptr) gives you the raw pointer you need and the docs say:
> The pointer is valid for as long as there are strong counts in the Arc.

I think in the cb you could use `ctx as *mut Arc<dyn Tls13SecretCallbacks>` as a ptr without needing the `Box`? Maybe? For IOCB we use `let io = unsafe { &*(ctx as *mut IOCB) };` to recover it.

wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1111,95 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+        id: ::std::os::raw::c_int,
+        secret: *const ::std::os::raw::c_uchar,
+        secret_size: ::std::os::raw::c_int,
+        ctx: *mut ::std::os::raw::c_void,
+    ) -> ::std::os::raw::c_int {
+        debug_assert!(!ssl.is_null());
+        debug_assert!(!secret.is_null());
+        debug_assert!(!ctx.is_null());
+
+        let keylog = Box::from_raw(ctx as *mut Arc<dyn Tls13SecretCallbacks>);
```
xv-ian-c, line 1155: I'm concerned what happens when this `keylog` goes out of scope the first time -- I think it would free the pointer to `Arc` which it contains? (But see above for a theory on how to avoid the box entirely).

wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1111,95 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+
+        let mut secret_keylog = match id as wolfssl_sys::Tls13Secret {
```
xv-ian-c, line 1157: It would be better to keep this as an enum IMHO. Ether by reexporting `Tls13Secret` or by providing a higher level wrapper for it.

wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1111,95 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+            wolfssl_sys::Tls13Secret_EARLY_EXPORTER_SECRET => "EARLY_EXPORTER_SECRET",
+            wolfssl_sys::Tls13Secret_EXPORTER_SECRET => "EXPORTER_SECRET",
+            _e => "UNKNOWN_SECRET",
+        }
+        .to_string();
+
+        secret_keylog.push(' ');
+
+        let mut random: Vec<u8> = vec![0; RANDOM_SIZE];
+        let random_size = if keylog.is_server() {
```
xv-ian-c, line 1176: Could we use `wolfSSL_is_server`?

xv-ian-c: Or can the app not just use `wolfSSL_get_*_random` itself indepdnent from this callback?

wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1111,95 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+                random.len(),
+            )
+        } else {
+            wolfssl_sys::wolfSSL_get_client_random(
+                ssl,
+                random.as_mut_ptr() as *mut c_uchar,
+                random.len(),
+            )
+        };
+        (0..random_size).for_each(|i| secret_keylog.push_str(&format!("{:02x}", random[i])));
```
xv-ian-c, line 1189: This feels like a `for i in random.iter() { ...}` loop or a `random.iter().foreach()` in a funny hat to me.

But as with the enum above why are we stringifying this stuff at this level? It would seem more natural if `Tls13SecretCallbacks::secret` took `which: Tls13Secret, random: &[u8], secret: &[u8]` as arguments?

xv-ian-c, line 1189: The C code handles `random_size <= 0` so I guess that is possible (and I suppose an error)?

kp-mariappan-ramasamy: Added two methods based on our discussion
One to get as a wireshark keyfile line
Another to get secret and random separately

wolfssl/src/ssl.rs
```diff
@@ -1176,8 +1168,9 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
         debug_assert!(!secret.is_null());
         debug_assert!(!ctx.is_null());
 
-        // let keylog = Box::from_raw(ctx as *mut Arc<dyn Tls13SecretCallbacks>);
-        let secret_cb = &*(ctx as *mut Tls13SecretCallbacksArg);
+        // SAFETY: We have only one strong reference in `self.secret_cb`
+        // Leak the Box pointer, so that it will not be freed while this function return
+        let secret_cb = Box::leak(Box::from_raw(ctx as *mut Tls13SecretCallbacksArg));
```
xv-ian-c, line 1173: Can you expand (here or perhaps in the commit message) on what the issue was with the previous code using `Arc::as_ptr` etc? I expected it would work and `secret_cb` here would have type `&Arc<Tls13SecretCallbacksArg>` which I thought would be fine.

Since this is obviously pretty subtle maybe `wolfssl/tests/ping_pong.rs` could register the cb when running in debug mode and `eprintln` it?

kp-mariappan-ramasamy: I do not know for sure.
This is my theory. I have tried some examples but could not see it 100%

But based on my reading, `Arc::as_ptr()` will return raw_pointer of data and in this case  its a `<dyn Callback>`. And it is a fat pointer, since its a trait object.

```rust
let secret_cb = Arc::as_ptr(&secret_cb) as *mut c_void;
```

And when we reconstruct it using :
```rust
let secret_cb =  &*(ctx as *mut Tls13SecretCallbacksArg);
```

We actually treat the pointer as the pointer to the Arc and not the actual data (trait object). So when it tries to dereference, and application crashes.

I have also tried `let secret_cb =  Arc::from_raw(ctx as *const dyn Tls13SecretCallbacks);`. But compiler does not like the type cast of `* mut c_void` to `*const dyn Tls13SecretCallbacks`.

So I come up with putting the Arc inside the box and send the box pointer as context pointer. It avoids the above fat pointer issue. While reconstruction, I just recreated a Box from context pointer and leak it at same time to avoid memory free at the end of the call.

wolfssl/src/debug.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Called when WolfSSL wishes to send new TLS1.3 secret used
+    /// `secret` is formatted as WireShark keylog string.
+    /// It can be saved to a file and used in WireShark directly
+    fn wireshark_keylog(&self, _secret: String) {}
+
+    /// Called when WolfSSL wishes to send new TLS1.3 secret used
+    /// Random value and secrets along with secret type is send separately
+    /// This can be used when application wants to just dump the secret/random
+    /// If the end-goal is to decrypt using Wireshark, use [`wireshark_keylog_line`] api
+    fn secrets(&self, _secret_type: Tls13Secret, _random: &[u8], _secret: &[u8]) {}
```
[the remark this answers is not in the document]
kp-mariappan-ramasamy: Oops. Thanks for correcting. I have updated the commit to reflect that.
Also code looks nicer now. Thanks.

wolfssl/src/debug.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::os::raw::{c_int, c_uint};
+use std::sync::Arc;
+
+/// Application provided callbacks to receive TLS1.3 secrets
+///
+/// There are two apis in this trait.
+/// Application can opt to receive either raw secrets or String which is suitable to use in
+/// WireShark keylog
+/// https://www.wireshark.org/docs/wsug_html_chunked/ChIOExportSection.html#ChIOExportTLSSessionKeys
+/// https://wiki.wireshark.org/TLS#using-the-pre-master-secret
```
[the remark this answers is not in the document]
kp-mariappan-ramasamy: That's because I hide it from command using feature flag.

wolfssl/src/ssl.rs
xv-ian-c: I'll have a play with this, leaking things should not be needed.

xv-ian-c: I figured it out and pushed it as a commit.

wolfssl/src/ssl.rs
```diff
@@ -1100,6 +1127,80 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
...
+        // containing the `Tls13Secretcallbacksarg` which is an `Arc`.
+        //
+        // We free `self.ssl` (the `wolfssl_sys::WOLFSSL`) on drop of
+        // `self`, any use of the callback must have stopped before
+        // the drop, since nothing can also be making calls into the session.
+        //
+        // Therefore `secret_cb` here is valid for as long as it needs to be.
+        let secret_cb: &mut Tls13SecretCallbacksArg = self.secret_cb.as_mut().unwrap();
+        let secret_cb: *mut Tls13SecretCallbacksArg = secret_cb as *mut Tls13SecretCallbacksArg;
+        let secret_cb = secret_cb as *mut c_void;
```
xv-ian-c, line 1147: Strictly speaking the type annotations here might not be needed and this could be a oneliner, but I think the types help clarify what is going on (and help the compiler alert us if something isn't doing what we thing).

I think the first annotation _might_ be needed since it causes auto deref to kick in, without that you'd get `&mut <Box<Arc<dyn ....>>>` not the type we want/need.

kp-mariappan-ramasamy: Looks nice. Thanks!

xv-ian-c (approved):

PR #120, opened 2024-03-27
CVPN-498: add tests for Chacha20Poly1305Aead cipher

kp-mariappan-ramasamy:
Added tests for Chacha20Poly1305Aead cipher

The test vector is taken from wolfssl repo with one change. Authtag was changed since we do not use AAD1 in our implementation


xv-thomas-leong (approved):

xv-andre-l:
> The test vector is taken from wolfssl repo

To help me and others understand this better, could you provide a permalink to the file containing this test vector?  I assume it's somewhere in https://github.com/wolfSSL/wolfssl, but I don't know where

kp-mariappan-ramasamy:
Sure updated the description and the commit message

expressvpn-andre-l (approved):

xv-ian-c (approved):
LGTM. Note that the ultimate source of these vectors appears to be https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-chacha20-poly1305-10#section-2.8.2.

PR #123, opened 2024-04-04
CVPN-753: Add public api for retrieving ErrorKind

kp-mariappan-ramasamy:
Add abstraction `FatalErrorKind` over wolfssl error code and add a public api for retrieving it

### DIFF
--- a/wolfssl/src/error.rs
+++ b/wolfssl/src/error.rs
@@ -2,6 +2,7 @@ use std::ffi::c_int;
 
 use bytes::Bytes;
 use thiserror::Error;
+use wolfssl_sys::wolfSSL_ErrorCodes_DOMAIN_NAME_MISMATCH as WOLFSSL_ERROR_DOMAIN_NAME_MISMATCH;
 
 /// The `Result::Ok` for a non-blocking operation.
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


wolfssl/src/error.rs
```diff
@@ -48,6 +57,37 @@ pub struct FatalError {
...
+#[derive(Clone, Copy, Debug)]
+pub enum FatalErrorKind {
+    /// Domain name mismatch error
+    DomainNameMismatch,
+    /// All other wolfssl fatal errors
+    Other(c_int),
+}
+
+impl From<FatalErrorKind> for FatalError {
+    #[inline]
```
xv-ian-c, line 80: I don't think we need to micromanage the compiler for this sort of thing...

wolfssl/src/error.rs
```diff
@@ -48,6 +57,37 @@ pub struct FatalError {
...
+/// Abstraction over WolfSSL errors
+#[derive(Clone, Copy, Debug)]
+pub enum FatalErrorKind {
+    /// Domain name mismatch error
+    DomainNameMismatch,
+    /// All other wolfssl fatal errors
+    Other(c_int),
+}
+
+impl From<FatalErrorKind> for FatalError {
```
xv-ian-c, line 79: What uses this? I can't spot a relevant use of `.into()`

wolfssl/src/lib.rs
```diff
@@ -18,7 +18,7 @@ pub use context::*;
 pub use rng::*;
 pub use ssl::*;
 
-pub use error::{Error, Poll, Result};
+pub use error::{Error, FatalError, FatalErrorKind, Poll, Result};
```
xv-ian-c, line 21: I wonder if `FatalErrorKind` should just be called `ErrorKind` and if we actually need both `FatalError`[^1] and `FatalErrorKind`?


WDYT of just replacing `FatalError` with `ErrorKind`:

```rust
pub enum ErrorKind {
    /// Domain name mismatch error
    DomainNameMismatch,
    /// All other wolfssl fatal errors
    Other{ what: String, code: c_int},
}
```
(note `Other` here now has the contents which `FatalError` has today).

Then Error is
```
pub enum Error {
    /// During secure renegotiation, if application data is found, we must call
    /// `wolfssl_read` to extract the data. If that `wolfssl_read` call fails,
    /// this error will be generated.
    #[error("App Data: {0}")]
    AppData(ErrorKInd),
    /// Top-level errors from WolfSSL API invocations.
    #[error("Fatal: {0}")]
    Fatal(ErrorKind),
}

impl Error {
   fn kind(&self) -> ErrorKind { ... }
}
```


[^1]: I think `FatalError` is actually a bad name, conceptually it's just the specific info about an error, it's no more or less fatal than any other error. If it weren't for the AppData vs other error use case it'd just be called `Error`. But that's a preexsiting issue arising from the whole AppData thing.

xv-ian-c (commented):
I had a thought on how to remove a layer from the API.

xv-ian-c (approved):

PR #126, opened 2024-04-11
Remove the `Mutex` from `WolfsslPointer` in `Session::ssl`

xv-ian-c:
Instead take `&mut self` in `Session`'s methods, this allows the consumers to choose to use an `&mut` session themselves and avoid locking, but does not preclude them using a `Mutex` themselves.

With this we now know that there can be no parallel code paths which might hit an `IOCallback` method, therefore we can make those receive `&mut self` (since the callback type is owned by the session).

Together with the use of `&mut self` in the various `Session` methods this now makes `io_cb_mut()` usable by consuming applications to access their IOCallback implementation's state knowing that nothing else can be touching it.

### DIFF
--- a/wolfssl/src/ssl.rs
+++ b/wolfssl/src/ssl.rs
@@ -6,7 +6,6 @@ use crate::{
 };
 
 use bytes::{Buf, Bytes, BytesMut};
-use parking_lot::Mutex;
 use thiserror::Error;
 
 use std::{
@@ -169,11 +168,9 @@ impl<IOCB: IOCallbacks> SessionConfig<IOCB> {
 // add traits such as `Send`.
 struct WolfsslPointer(NonNull<wolfssl_sys::WOLFSSL>);
 
-impl std::ops::Deref for WolfsslPointer {
-    type Target = NonNull<wolfssl_sys::WOLFSSL>;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
+impl WolfsslPointer {
+    fn as_ptr(&mut self) -> *mut wolfssl_sys::WOLFSSL {
+        self.0.as_ptr()
     }
 }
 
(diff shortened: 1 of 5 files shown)


wolfssl/src/ssl.rs
```diff
@@ -922,7 +890,10 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
         // owned by the `Session`. See `register_io_context` below for
         // an argument as to why IO will be stopped (by releasing
         // `WOLFSSL`) before that box is dropped.
-        let io = unsafe { &*(ctx as *mut IOCB) };
+        //
+        // Since `io` is owned by the `Session` and we know that `try_read` takes the lock.
+        // XXX needs thought since lock doesn't protect io
```
kp-mariappan-ramasamy, line 895: I think this case is fine, since now `Session::try_read` take mutable Session when in turn calls `wolfSSl_read` -> `io_recv_shim`
So at this point, io_recv_shim will have ctx aka. Session.ssl will have exclusive access and hence safe to access.

Please let me know if my understanding is wrong. If not, please update the comment to reflect the same.

kp-mariappan-ramasamy (commented):
LGTM except a question about a comment

wolfssl/src/ssl.rs
xv-ian-c: I think you are correct. IIRC I wrote this before I'd worked out the whole `&mut self` change.

xv-ian-c: Updated the comment (and in the write path too)

wolfssl/src/ssl.rs
```diff
@@ -950,7 +923,12 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
         // owned by the `Session`. See `register_io_context` below for
         // an argument as to why IO will be stopped (by releasing
         // `WOLFSSL`) before that box is dropped.
-        let io = unsafe { &*(ctx as *mut IOCB) };
+        //
+        // The `io` (type: `IOCB`) is owned by the `Session` and we
+        // know that `try_read` holds an exclusive (`&mut`) reference
```
kp-mariappan-ramasamy, line 928:
```suggestion
        // know that `try_write` holds an exclusive (`&mut`) reference
```

xv-ian-c: Good spot!

kp-mariappan-ramasamy (approved):

PR #131, opened 2024-04-26
CVPN-249 Add Release Automation

xv-thomas-leong:
This PR is adding a GitHub Workflow to automate crate release in this repository

## Description

- The release automation has 2 pre-conditions
     1. Crates' Cargo.toml is changed
     2. The PR has the `release` label
 - If the above were satisified, the automation will be triggered
 - The automation will run in a "dry-run" mode until the PR is merged
 - When doing a non-dry-run, i.e. production release, the automation
    will create a release on both Crate.io and GitHub

## How Has This Been Tested?

CI tested.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] The correct base branch is being used, if not `main`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,167 @@
+name: Release
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+    paths:
+      - "*/Cargo.toml" # only run if Cargo.toml changes
+  push:
+    branches: [main]
+    paths:
+      - "*/Cargo.toml" # only run if Cargo.toml changes
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref_name != 'main' }}
+
+jobs:
+  check_label:
+    runs-on: ubuntu-latest
+    outputs:
+      is_release_label_attached: ${{ steps.pr_labels.outputs.is_release_label_attached }}
+      issue_number: ${{ steps.pr_labels.outputs.issue_number }}
+    steps:
+      - uses: jwalton/gh-find-current-pr@v1
+        id: find_pr
+        with:
+          state: all
+      - id: pr_labels
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const issue_number = ${{ steps.find_pr.outputs.pr }}
+            const owner = context.repo.owner
+            const repo = context.repo.repo
+
+            const attached_labels = (await github.rest.issues.listLabelsOnIssue({ owner, repo, issue_number })).data
+
+            const is_release_label_attached = attached_labels.find(label => label.name == 'release') !== undefined
+            core.setOutput('is_release_label_attached', is_release_label_attached)
+            core.setOutput('issue_number', issue_number)
+  release:
+    runs-on: ubuntu-latest
+    needs: check_label
+    if: needs.check_label.outputs.is_release_label_attached == 'true'
+    env:
+      EARTHLY_TOKEN: "${{ secrets.EARTHLY_TOKEN }}"
+      CARGO_REGISTRY_TOKEN: "${{ secrets.CARGO_REGISTRY_TOKEN }}"
+      FORCE_COLOR: 1
+    steps:
+      - uses: earthly/actions-setup@v1
+        with:
+          version: v0.8.3
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+      - run: npm install toml globby axios
+      - uses: actions/github-script@v7
+        id: process_release
+        with:
+          script: |
+            const is_merge_event = ${{ github.event_name == 'push' }}
+
+            const { globby } = await import("${{ github.workspace }}/node_modules/globby/index.js");
+            const toml = require('toml');
+            const fs = require('fs');
+            const axios = require('axios');
+            const { spawn } = require('child_process');
+
+            async function get_crate_details(name) {
+              try {
+                return (await axios(`https://crates.io/api/v1/crates/${name}`)).data
+              } catch(error) {
+                console.error("Error fetching crate details:", error)
+              }
+            }
+
+            function run_command_stream(command_arr) {
+              const [command, ...args] = command_arr
+
+              const child = spawn(command, args)
+              child.stdout.on("data", data => console.log(Buffer.from(data).toString()))
+              child.stderr.on("data", data => console.error(Buffer.from(data).toString()))
+              child.on("close", code => {
+                console.log(command, "exited with code", code)
+                if (code != 0) {
+                  core.setFailed(`Command ${command_str} exited with code ${code}`)
+                }
+              })
+            }
+
+            async function create_github_release(crate_name, version) {
+              console.log("Creating a github release for", crate_name, version)
+
+              const { owner, repo } = context.repo
+              const tag_name = `${crate_name}-${version}`
+              const name = tag_name
+              const ref = `refs/tags/${tag_name}`
+              const sha = context.sha
+              const body = ""
+
+              await github.rest.git.createRef({ owner, repo, ref, sha })
+              await github.rest.repos.createRelease({ owner, repo, tag_name, name, body })
+            }
+
+            const is_dry_run = !is_merge_event
+            console.log("Dry run status is:", is_dry_run)
+
+            const crate_definitions = await globby("*/Cargo.toml")
+
+            const crates_to_publish = []
+            for (crate_definition of crate_definitions) {
+              console.log(`Processing ${crate_definition}`)
+
+              const data = toml.parse(fs.readFileSync(crate_definition, 'utf8'))
+              const name = data.package.name
+              const version_on_file = data.package.version
+              const version_on_crate_io = (await get_crate_details(name)).crate.max_version
+
+              if (version_on_file != version_on_crate_io) {
+                console.log("Crate", name, "need publishing")
+                crates_to_publish.push({ name, version_on_file, version_on_crate_io })
+
+                run_command_stream(["earthly", "--push", "--ci", "--org", "expressvpn", "--satellite", "wolfssl-rs", "--secret", "CARGO_REGISTRY_TOKEN", "+publish", `--PACKAGE=${name}`, `--DRY_RUN=${is_dry_run}`])
+                if (!is_dry_run) {
+                  await create_github_release(name, version_on_file)
+                }
+              } else {
+                console.log("Crate", name, "does not need publishing")
+              }
+            }
+
+            const crates_to_publish_table_rendered = crates_to_publish.map(item =>
+              `| \`${item.name}\` | ${item.version_on_crate_io} | **${item.version_on_file}** |`
+            ).join("\n")
+            core.setOutput('crates_to_publish_table_rendered', crates_to_publish_table_rendered)
+
+      - uses: peter-evans/find-comment@v3
+        id: fc
+        with:
+          issue-number: ${{ needs.check_label.outputs.issue_number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Release Plan
+      - if: steps.process_release.outputs.crates_to_publish_table_rendered == ''
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ needs.check_label.outputs.issue_number }}
+          body: |
+            # :rocket: Release Plan
+
+            Nothing to release
+          edit-mode: replace
+      - if: steps.process_release.outputs.crates_to_publish_table_rendered != ''
+        uses: peter-evans/create-or-update-comment@v4
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ needs.check_label.outputs.issue_number }}
+          body: |
+            # :rocket: Release Plan
+
+            | Crate | Previous version | New version |
+            |--|--|--|
+            ${{ steps.process_release.outputs.crates_to_publish_table_rendered }}
+          edit-mode: replace
--- a/Earthfile
+++ b/Earthfile
@@ -79,3 +79,17 @@ fmt:
 check-dependencies:
     FROM +copy-src
     DO lib-rust+CARGO --args="deny --all-features check --deny warnings bans license sources"
+
+# publish publishes the target crate to cargo.io. Must specify package by --PACKAGE=<package-name>
+publish:
+    FROM +copy-src
+    ARG --required PACKAGE
+    ARG DRY_RUN=true
+
+    LET DRY_RUN_OPTION=""
+    IF [ "$DRY_RUN" = "true" ]
+        SET DRY_RUN_OPTION="--dry-run"
+    END
+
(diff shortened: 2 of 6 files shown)


Earthfile
```diff
@@ -79,3 +79,20 @@ fmt:
...
+
+# publish publishes the target crate to cargo.io. Must specify package by --PACKAGE=<package-name>
+publish:
+    FROM +copy-src
+    ARG --required PACKAGE
+    ARG DRY_RUN=true
+
+    # earthly doesn't support pass secrets to FUNCTION directly, copy it to env through a file
+    RUN --secret CARGO_REGISTRY_TOKEN echo $CARGO_REGISTRY_TOKEN > ~/.CARGO_REGISTRY_TOKEN
+    ENV CARGO_REGISTRY_TOKEN=$(cat ~/.CARGO_REGISTRY_TOKEN)
```
xv-ian-c, line 91: https://docs.earthly.dev/docs/earthfile#env-same-as-dockerfile-env says:
> Do not use the ENV command for secrets used during the build. All ENV values used during the build are persisted within the image itself

Maybe it's ok here since we aren't actually making a final image.

But `DO lib-rust+CARGO` is mainly about caching the `target` dir for incremental builds -- is that useful `cargo publish`? I suspect not so maybe we can just do `RUN --secret cargo publish...`?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,152 @@
+name: Release
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+    paths:
+      - "*/Cargo.toml" # only run if Cargo.toml changes
+  pull_request_target:
+    types: [closed]
```
xv-ian-c, line 9: Why do we want to run when a PR is closed?

xv-ian-c: I think we more likely want:
```yaml
on:
  pull_request:
  push:
    branches:
      - main
```

So we run on:
* Any PR (no need for types, we want them all I think)
* Any push to main (== PR merged)

Could make it
```yaml
on:
  pull_request:
    branches:
      - main
  push:
    branches:
      - main
```
which would only run on PRs directly targeting main (so not stacked PRs).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,152 @@
...
+              const data = toml.parse(fs.readFileSync(crate_definition, 'utf8'))
+              const name = data.package.name
+              const version_on_file = data.package.version
+              const version_on_crate_io = (await get_crate_details(name)).crate.max_version
+
+              if (version_on_file != version_on_crate_io) {
+                console.log("Crate", name, "need publishing")
+                crates_to_publish.push({ name, version_on_file, version_on_crate_io })
+
+                run_command_stream(`earthly -ci --org expressvpn --satellite wolfssl-rs --secret CARGO_REGISTRY_TOKEN +publish --PACKAGE=${name} --DRY_RUN=${is_dry_run}`)
```
xv-ian-c, line 124: `-ci` should be `--ci`?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,152 @@
...
+            async function get_crate_details(name) {
+              try {
+                return (await axios(`https://crates.io/api/v1/crates/${name}`)).data
+              } catch(error) {
+                console.error("Error fetching crate details:", error)
+              }
+            }
+
+            function run_command_stream(command_str) {
+              const [command, ...args] = command_str.split(" ")
```
xv-ian-c, line 57: danger of shell quoting problems here. would it be so bad to have to call `run_command_stream(["foo", "bar"])`?

Earthfile
xv-thomas-leong: I think in this scenario we care more about protecting secrets than caching. I agree we should switch to `RUN --secret cargo publish...` 👍

.github/workflows/release.yaml
xv-thomas-leong: For your questions:

> Why do we want to run when a PR is closed?

Normally we can use `on: push`, but it will set the actions context to a branch instead of a PR, which would make it hard (but not impossible) for me to look up PR information, such as labelling. In my design, I check for whether the label `release` is attached to the PR such that we can control whether we are intentionally trying to release something.

I have done `look up PR information from the main branch after merge` in the past, but the solution is very janky. I was under different constraints back then, when that was the only viable solution. I would not suggest we do that here.

`pull_request_target` is a nice alternative because it has the context of a PR but runs on the base branch (i.e. the SHA of the commit would be in the main branch, instead of the last commit in a PR). And I have a guard to only disable dry-run when the PR is merged.

---

> Any PR (no need for types, we want them all I think)

The [GitHub Actions docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) says:

```
By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened
```

Which means I won't get labeled/unlabeled events by default, so I have to specify them manually.

---

> which would only run on PRs directly targeting main

With my current `pull_request_target` config, the action would still run when someone closes the PR without merging. But because of my dry-run check, it will run in the dry-run mode which is harmless.

.github/workflows/release.yaml
xv-thomas-leong: Oh yes, thank you for catching that, will fix it 🙏

.github/workflows/release.yaml
xv-thomas-leong: Good point. Let me change 🙏

.github/workflows/release.yaml
xv-ian-c: > , which would make it hard (but not impossible) for me to look up PR information, such as labelling

https://github.com/jwalton/gh-find-current-pr could be used? We use successfully in `.github/workflows/ci.yaml`.

> pull_request_target is a nice alternative because it has the context of a PR but runs on the base branch (i.e. the SHA of the commit would be in the main branch, instead of the last commit in a PR). And I have a guard to only disable dry-run when the PR is merged.

So on a PR we run on main and not on the PR? That seems odd on first glance.

I think maybe adding some "How to do a release" documentation might help clarify things and also be useful for anyone who comes to make a release in the future.


> Which means I won't get labeled/unlabeled events by default, so I have to specify them manually.

Thanks, I hadn't realized about the release label stuff, BTW In the commit message you said "release tag" (if I'd thought about it I'd have realized that was odd and you probably meant label, but I didn't...)

xv-thomas-leong: > https://github.com/jwalton/gh-find-current-pr could be used?

I think I have implemented `jwalton/gh-find-current-pr` on my own before, that is exactly the janky solution I was talking about. Although I believe `pull_request_target` is the right way of getting PR context, `on: push: branches: [main]`is more intuitive which would prevent us from making mistakes in the future. Let me make changes to that.

> I think maybe adding some "How to do a release" documentation

Adding documentation on how to do a release is definitely on my TODO list. And I will make sure to add PR descriptions I make it ready to review. Btw thanks for the early review!

> In the commit message you said "release tag" 

Let me update the commit messages. I must have mixed up with the wordings.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,177 @@
...
+              const comment_header = `<!-- ${identifier} -->\n`
+              const comment = comment_header + comment_body
+
+              const issue_number = ${{ steps.find_pr.outputs.pr }}
+              const owner = context.repo.owner
+              const repo = context.repo.repo
+
+              const existing_comments = (await github.rest.issues.listComments({
+                  issue_number, owner, repo
+              })).data
```
xv-ian-c, line 105: FYI you could use peter-evans/find-comment@v3 and peter-evans/create-or-update-comment@v4 to replace a bunch of this.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,177 @@
...
+              const version_on_crate_io = (await get_crate_details(name)).crate.max_version
+
+              if (version_on_file != version_on_crate_io) {
+                console.log("Crate", name, "need publishing")
+                crates_to_publish.push({ name, version_on_file, version_on_crate_io })
+
+                run_command_stream(["earthly", "--ci", "--org", "expressvpn", "--satellite", "wolfssl-rs", "--secret", "CARGO_REGISTRY_TOKEN", "+publish", `--PACKAGE=${name}`, `--DRY_RUN=${is_dry_run}`])
+                if (!is_dry_run) {
+                  await create_github_release(name, version_on_file)
+                }
```
xv-ian-c, line 152: This creates a release for each crate, so if multiple crates are included in the ` for (crate_definition of crate_definitions) {` above, what happens?  i.e. what will https://github.com/expressvpn/wolfssl-rs/releases look like, will it be confusing wrt the different crates with different versions interleaved? If we release package A ten times after the last package B release will package  B's last release disappear off the end?

Do we actually want/need a GH release here, or would just a tag do?

README.md
```diff
@@ -58,4 +58,16 @@ Please refer to [official documentation for Earthly Satellites](https://docs.ear
...
+A GitHub Workflow is set up to automate the release of crates in this repo. Upon a release, it will create a release in GitHub and Crates.io
+
+To create a new release, follow the below steps:
+
+1. Bump the version in `<crate-name>/Cargo.toml`. We follow the semantic versioning pattern when deciding a new version number
+1. Open a PR, attach the `release` label to the PR
+1. Observe that a comment is add to the PR, indicating the current version and the upcoming version
+1. Merge the PR, a new version should be released to both GitHub and Crates.io
+
+The automation supports releasing multiple crates in one PR. It can be done by bumping the version in multiple `Cargo.toml` files.
```
xv-ian-c, line 73: This might be tricky if/when wolfssl has a versioned (rather than path) dependency on wolfssl-sys? Depending what checks `cargo publish` does...

.github/workflows/release.yaml
```diff
@@ -0,0 +1,177 @@
...
+                  core.setFailed(`Command ${command_str} exited with code ${code}`)
+                }
+              })
+            }
+
+            async function create_or_update_comment(identifier, comment_body) {
+              const comment_header = `<!-- ${identifier} -->\n`
+              const comment = comment_header + comment_body
+
+              const issue_number = ${{ steps.find_pr.outputs.pr }}
```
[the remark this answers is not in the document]
xv-thomas-leong: I think that depends on whether we care more about avoiding repetition or separation of concerns here. 

In this case, I don't think repeating the `find_pr` step is too concerning, as the code complexity and run-time cost are low. With that, I would want to keep the `check_label` step doing just one thing - check for labels.

What do you think?

.github/workflows/release.yaml
xv-thomas-leong: For logic that is as simple as this one, how should we choose whether to use a library or implement it ourselves?

.github/workflows/release.yaml
xv-thomas-leong: Package B's last release will disappear off the end, but it is fine as it will still be searchable. 

For your question "would just a tag do?", if the purpose is to retrieve the commit based on version number, then yes, a tag would do. But GitHub release seems to do more than just that:

1. Show up on people's feed when people "follow" a repository:

image

2. Allow people to easily compare between releases

https://docs.github.com/en/repositories/releasing-projects-on-github/comparing-releases

.github/workflows/release.yaml
xv-ian-c: I generally lean towards using exising/debugged libraries, at least if they seem reasonably widely used.

README.md
xv-thomas-leong: Updated the release instruction that the two crates should be released together, but separately.

Earthfile
```diff
@@ -79,3 +79,17 @@ fmt:
...
+    ARG --required PACKAGE
+    ARG DRY_RUN=true
+
+    LET DRY_RUN_OPTION=""
+    IF [ "$DRY_RUN" = "true" ]
+        SET DRY_RUN_OPTION="--dry-run"
+    END
+
+    # earthly doesn't support passing secrets to FUNCTION directly. Calling the cargo command directly instead.
+    RUN --secret CARGO_REGISTRY_TOKEN cargo publish --package $PACKAGE $DRY_RUN_OPTION
```
kp-mariappan-ramasamy, line 95: We might add `--push` option to make sure, we don't run it locallyhttps://docs.earthly.dev/docs/earthfile#options-1

.github/workflows/release.yaml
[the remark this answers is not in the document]
kp-mariappan-ramasamy: To me, it looks outputting the PR number from check_label makes sense, since  `check_label` job's label_attached output is specific to a PR.
label_attached without PR number lacks context.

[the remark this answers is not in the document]
xv-thomas-leong: Addressed in 3d26d56

.github/workflows/release.yaml
xv-thomas-leong: Updated to use the pre-built actions in 3d26d56

Earthfile
xv-thomas-leong: Added in 3d26d56

xv-ian-c (commented):
LGTM, one minor comment.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,167 @@
...
+    paths:
+      - "*/Cargo.toml" # only run if Cargo.toml changes
+  push:
+    branches: [main]
+    paths:
+      - "*/Cargo.toml" # only run if Cargo.toml changes
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
xv-thomas-leong: Let me change to `cancel-in-progress: ${{ github.ref_name != 'main' }}` to be on the safe side.

README.md
```diff
@@ -58,4 +58,22 @@ Please refer to [official documentation for Earthly Satellites](https://docs.ear
...
 If you are not a member of ExpressVPN, you may set up your own Earthly satellite according the official instructions above.
- 
\ No newline at end of file
+
+# Releasing
+
+This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. They should __always__ be released together. Since `wolfssl` depends on `wolfssl-sys`, they should be released in the below order:
+
+1. Bump the version in `wolfssl-sys` and release it (Follow the section [Releasing a Single Crate](#releasing-a-single-crate))
+1. Bump the version in `wolfssl`, update the dependency `wolfssl-sys` to the latest version, and release it (same procedure as above)
```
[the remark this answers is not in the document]
xv-thomas-leong: No, I do not intend to release the `wolfssl` crate in this PR.

.github/workflows/release.yaml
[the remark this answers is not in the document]
xv-ian-c: Yes, good idea

README.md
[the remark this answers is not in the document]
xv-ian-c: I don't think making that change involves actually releasing, it preparing the ground for when we do release in the future. If we don't do it now then we need to remember to do it the first time we do release.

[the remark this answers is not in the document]
xv-thomas-leong: Added in 8dca837

xv-ian-c (approved):

PR #147, opened 2024-06-18
GHA: Coordinate wolfssl and wolfssl-sys releases

xv-ian-c:
When releasing `wolfssl` from `main` we want to be sure it depends on the most recent `wolfssl-sys` changes in `main`.

For usage from git this is fine since the dependency involves a `path` key. This is also what we actually test in CI.

For usage from crates.io this means we want the `version` field to point to the latest version of the code -- which may mean we need to release `wolfssl-sys` first if there are unreleased changes there.

The SHA of the tree object of the `wolfssl-sys` subtree changes precisely when `wolfssl-sys` is changed, so we can match using that by comparing main to the tagged release.

There are four states:
* `released`: `wolfssl` depends on a release of `wolfssl-sys` which contains the most recent code -> go ahead and release
* `out-of-date`: `wolfssl` depends on a release of `wolfssl-sys` which is not the most recent -> block the release
* `pending`: `wolfssl` depends on a release of `wolfssl-sys` which has not been released (meaning the git tag corresponding to `wolfssl-sys/Cargo.toml` version does not exist) -> block the release
* `unreleased`: `wolfssl` depends on latest release of `wolfssl-sys` but there are changes in git since that release was made (git tag corresponding to the version is behind latest code in the `wolfssl-sys` subtree) -> block the release

These are reflected in the `dependency_status` output.

I think if we could control the order in which the `process_release` step considers the crates so that it finds (and releases) `wolfssl-sys` before `wolfssl` then the `pending` result would be non-blocking, and we could potentially release both crates with a single PR.  instead of one by one. I think we should land this change first and ensure that it works though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,15 +68,27 @@ If you are not a member of ExpressVPN, you may set up your own Earthly satellite
 
 This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. It is required when releasing `wolfssl` that it depends on an version of `wolfssl-sys` which is up to date with `main`.
 
-If `wolfssl-sys` has unreleased changes:
+There are four possible states for `wolfssl`'s dependency on `wolfssl-sys`:
 
-1. Bump the crate version `wolfssl-sys`
+| State         | Description                                                               |
+|---------------|---------------------------------------------------------------------------|
+| `released`    | `wolfssl` depends on the latest, most up to date release of `wolfssl-sys` |
+| `pending`     | `wolfssl` depends on a pending release of `wolfssl-sys`                   |
+| `unreleased`  | `wolfssl-sys` has changes which must be released                          |
+| `out-of-date` | `wolfssl` depends on an old version of `wolfssl-sys`                      |
+
+If state is `out-of-date` or `unreleased` then:
+
+1. Bump the crate version in `wolfssl-sys/Cargo.toml`
 1. Update the version specified under `dependencies` in the `wolfssl` crate
+
+The state will now be `pending` (or it already was) so:
+
 1. Release `wolfssl-sys` (Follow the section [Releasing a Single Crate](#releasing-a-single-crate))
 
 Once the most recent `wolfssl-sys` is released and `wolfssl` depends on it then `wolfssl` can be released:
 
-1. Bump the crate version `wolfssl`
+1. Bump the crate version in `wolfssl/Cargo.toml`
 1. Release `wolfssl` (Follow the section [Releasing a Single Crate](#releasing-a-single-crate))
 
 ## Releasing a Single Crate
--- a/README.md
+++ b/README.md
@@ -66,11 +66,18 @@ If you are not a member of ExpressVPN, you may set up your own Earthly satellite
 
 # Releasing
 
-This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. They should __always__ be released together. Since `wolfssl` depends on `wolfssl-sys`, they should be released in the below order:
+This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. It is required when releasing `wolfssl` that it depends on an version of `wolfssl-sys` which is up to date with `main`.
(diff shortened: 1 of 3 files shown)


xv-ian-c:
After testing I've removed the release label as well as the commit which bumped things.

README.md
```diff
@@ -66,11 +66,18 @@ If you are not a member of ExpressVPN, you may set up your own Earthly satellite
...
-1. Bump the crate version `wolfssl` and release it (same procedure as above)
+If `wolfssl-sys` has unreleased changes:
 
+1. bump the crate version `wolfssl-sys`
+1. update the version specified under `dependencies` in the `wolfssl` crate
+1. release `wolfssl-sys` (Follow the section [Releasing a Single Crate](#releasing-a-single-crate))
+
+Once the most recent `wolfssl-sys` is released and `wolfssl` depends on it then `wolfssl` can be released:
+
+1. Bump the crate version `wolfssl`
```
kp-mariappan-ramasamy, line 79: nit: inconsistent case usage

.github/workflows/check-dependencies.yaml
```diff
@@ -0,0 +1,110 @@
...
+        run: |
+          TAG="wolfssl-sys-${{ steps.get-sys-release.outputs.version }}"
+          if ! git fetch origin --depth=1 tag "$TAG" ; then
+              echo "tag $TAG not found, assuming pending release"
+              echo sha=pending >> $GITHUB_OUTPUT
+              exit 0
+          fi
+
+          TREESHA=$(git ls-tree --object-only $TAG wolfssl-sys)
+          echo TREESHA=$TREESHA
```
kp-mariappan-ramasamy, line 49: FYI, we could use github step/job summary to print these things.
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

Might be easier to take a look for these release workflows

kp-mariappan-ramasamy (approved):
LGTM minor nits

@xv-thomas-leong Think you might be interested in this changes

.github/workflows/check-dependencies.yaml
xv-ian-c: Ooh, nice, let me have a play with that!

xv-ian-c: They appear in an overall job summary, like https://github.com/expressvpn/wolfssl-rs/actions/runs/9594333419.

Which is nice but I think we might want the per-step echoes when debugging too?

README.md
```diff
@@ -66,11 +66,18 @@ If you are not a member of ExpressVPN, you may set up your own Earthly satellite
 
 # Releasing
 
-This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. They should __always__ be released together. Since `wolfssl` depends on `wolfssl-sys`, they should be released in the below order:
+This repository is a monorepo for two crates: `wolfssl-sys` and `wolfssl`. It is required when releasing `wolfssl` that it depends on an version of `wolfssl-sys` which is up to date with `main`.
 
-1. Bump the crate version `wolfssl-sys`, update the version specified under `dependencies` in the `wolfssl` crate, and release it (Follow the section [Releasing a Single Crate](#releasing-a-single-crate))
-1. Bump the crate version `wolfssl` and release it (same procedure as above)
+If `wolfssl-sys` has unreleased changes:
```
kp-stanislav-baiduzhyi, line 71: Might be useful to copy over the explanation of states from the PR description into here as well.

.github/workflows/check-dependencies.yaml
xv-ian-c: https://github.com/expressvpn/wolfssl-rs/actions/runs/9594476612 is what I ended up with, basically the output of the two `validate` steps goes there but the various probes log with echo.

xv-thomas-leong (approved):

kp-mariappan-ramasamy (approved):

PR #150, opened 2024-06-24
Release wolfssl v1.1.0

xv-ian-c:
Following #148 release the `wolfssl` crate too.

PR #151, opened 2024-06-24
Release with fresh `Cargo.lock`

xv-ian-c:
#148 and #150 missed updating `Cargo.lock` do so and add a CI step so it cannot be forgotten in the future

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,6 +56,20 @@ jobs:
         with:
           submodules: true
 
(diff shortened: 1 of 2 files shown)


xv-raihaan-m (approved):
Does this mean we have to fix the releases?

xv-ian-c:
Luckily not, `Cargo.lock` is a repo-local thing which isn't part of the release to crates.io. Releases only care about the constraints from `*/Cargo.toml`.

PR #152, opened 2024-06-24
Release both crates via a single release PR

xv-ian-c:
If the `wolfssl-sys` release is pending then by ensuring we process that release first we can also release `wolfssl` in the same workflow.
    
Updated the documentation and add a flowchart to try and clarify the needed steps.

[Rendered doc](https://github.com/expressvpn/wolfssl-rs/blob/3a8c664fc234dc2b35cb466945fcbc9474a11281/README.md) .

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -133,11 +133,13 @@ jobs:
             console.log("wolfssl to sys dependency version: ${{ steps.check_dependencies.outputs.dependency_version }}")
             console.log("wolfssl to sys dependency status: ${{ steps.check_dependencies.outputs.dependency_status }}")
 
-            const crate_definitions = await globby("*/Cargo.toml")
+            const crates = ["wolfssl-sys", "wolfssl"]
 
             let blocked_crates = 0
             const crates_to_publish = []
-            for (crate_definition of crate_definitions) {
+            for (crate of crates) {
+              const crate_definition = crate + "/Cargo.toml"
+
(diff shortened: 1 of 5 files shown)


xv-ian-c:
This doesn't work. The dry run release of wolfssl doesn't complete because the pending wolfssl-sys doesn't exist yet:

```
            +publish *failed* | Caused by:
            +publish *failed* |   failed to select a version for the requirement `wolfssl-sys = "^1.1.1"`
            +publish *failed* |   candidate versions found which didn't match: 1.1.0, 1.0.0, 0.1.16, ...
            +publish *failed* |   location searched: crates.io index
            +publish *failed* |   required by package `wolfssl v1.1.1 (/wolfssl-rs/target/package/wolfssl-1.1.1)`
            +publish *failed* |   perhaps a crate was updated and forgotten to be re-vendored?
            +publish *failed* | ERROR Earthfile line 95:4
            +publish *failed* |       The command
            +publish *failed* |           RUN --push cargo publish --package $PACKAGE $DRY_RUN_OPTION
            +publish *failed* |       did not complete successfully. Exit code 101
```

In theory we could add `--no-verify` to the `cargo publish --dry-run` but that likely turns off other checks which are likely to be useful.

xv-ian-c:
I did look into using --no-verify but it does cover useful looking stuff which our other build and test jobs do not.

So this seems like a non-starter. I rescued the diagram into #153.

PR #153, opened 2024-06-24
README: Add diagram to clarify release flow.

xv-ian-c:
Also make it clearer that if there are no wolfssl-sys changes (state is `released` already) then wolfssl can be released by itself.

---

[Rendered](https://github.com/expressvpn/wolfssl-rs/blob/afa35c4f904a1f02c0cc297b9cbb9beb2b71edc7/README.md#releasing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ There are four possible states for `wolfssl`'s dependency on `wolfssl-sys`:
 | `unreleased`  | `wolfssl-sys` has changes which must be released                          |
 | `out-of-date` | `wolfssl` depends on an old version of `wolfssl-sys`                      |
 
+If state is `released` then `wolfssl` can be released without releasing a new version of `wolfssl-sys`.
+
 If state is `out-of-date` or `unreleased` then:
 
 1. Bump the crate version in `wolfssl-sys/Cargo.toml`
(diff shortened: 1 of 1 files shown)


kp-mariappan-ramasamy:
Somehow it feels more confusing with the diagram :sweat: 

I think its most the out-of-date flow which seems complex:

Out-of-date -->
Update dependency from wolfssl on wolfssl-sys in wolfssl/Cargo.toml to new version --> 
Pending -->
Raise, review and merge wolfssl-sys release PR -->
Released -->
Update version in wolfssl/Cargo.toml -->

Why do we need to update version in wolfssl/Cargo.toml twice. Does not outdated means, wolfssl, depends on old wolfssl-sys and update wolfssl/Cargo alone should fix that ???

xv-ian-c:
In the first update to `wolfssl/Cargo.toml` we are changing the:
```toml
wolfssl-sys = { path = "../wolfssl-sys", version = "1.1.0" }
```
line to use the new `wolfssl-sys` version.

In the second we are changing:
```
version = "1.1.0"
```
to release a new version of wolfssl.

In theory we could defer the first step and do it in the second, IOW have wolfssl depend on the previous release for a window of time, I'm not sure if that would cause issues though..

xv-ian-c:
I tried to clarify the diagram a bit.

kp-mariappan-ramasamy (approved):

PR #157, opened 2024-07-08
build: only include oqs features we use

xv-raihaan-m:
This reduces the size of the .rlib file by 2 MB in release. Tested with the earthly +build-release.

One thing though, I don't think we even use dilithium or falcon at this stage but lightway-core includes them. So I'll leave it in for now.


xv-ian-c (approved):
LGTM.

PR #162, opened 2024-07-23
Use correct *runtime* checks for (D)TLS 1.3

xv-ian-c:
The `Protocol` type corresponds to a request to WolfSSL at setup time to take on a particular role (client vs server) supporting some set of versions (DTLS, TLS, 1.x etc). Accordingly it contains some "umbrella" versions such as
`DtlsClient` and `TlsServer` which request only "some version of" the respective protocols.
    
Therefore it does not make sense to ask "is this protocol DTLS 1.3 etc" since that would be false for `DtlsClient` even though the protocol which has actually been negotiated may well be version 1.3.
    
Move the helper methods to the `ProtocolVersion` type which reflects the runtime negotiated actual version and make use of `Session::version()` in the places which previously used the `protocol` field (which is now unneeded).

After having done this rename `Protocol` to `Method` (which is what WolfSSL actually calls the corresponding thing), in the hopes of making it harder to make the same mistake in the future.

### DIFF
--- a/wolfssl/src/context.rs
+++ b/wolfssl/src/context.rs
@@ -2,7 +2,7 @@ use crate::{
     callback::IOCallbacks,
     error::{Error, Result},
     ssl::{Session, SessionConfig},
-    CurveGroup, NewSessionError, Protocol, RootCertificate, Secret, SslVerifyMode,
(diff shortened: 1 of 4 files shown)


xv-raihaan-m (approved):
LGTM! Thanks for fixing the bug

kp-mariappan-ramasamy (approved):
LGTM

PR #166, opened 2024-07-26
GHA: Ensure that the CI nexus job fails if any dependent job fails

xv-ian-c:
This is required to have it actually block merging PRs if one of the constituent jobs fails, since such a failure will mark the `ci` job as "skipped" which does not block merging.

See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,10 +12,11 @@ jobs:
   ci:
     needs: [earthly, coverage]
     runs-on: ubuntu-latest
+    if: always()
     steps:
       - shell: bash
         run: |
-          echo "Build success"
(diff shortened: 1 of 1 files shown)


xv-ian-c:
NB Just adding always() is not sufficient, since it causes the ci job to run and pass even if a prerequisite failed. I also tried the more complex condition from https://stackoverflow.com/questions/69354003/github-action-job-fire-when-previous-job-skipped.

After testing on other repositories I settled on validating `toJSON(needs)` which is clearer/more direct.

kp-mariappan-ramasamy (approved):
LGTM

PR #192, opened 2024-11-07
CVPN-1554 Remove liboqs and use WolfSSL's implementations 

kp-thomas-yau:
# Description
This PR removes `liboqs` and would instead use WolfSSL's implementation of Kyber. Since WolfSSL would officially release their Kyber/ML-KEM implementations a few months later, we would use the git patch to essentially patch their PRs on top of the 5.7.4 release.

The patch consists of the commits and code changes from the following PR from WolfSSL:
- https://github.com/wolfSSL/wolfssl/pull/8143
- https://github.com/wolfSSL/wolfssl/pull/8172
- https://github.com/wolfSSL/wolfssl/pull/8183
- https://github.com/wolfSSL/wolfssl/pull/8185

Configuration for enabling ML-KEM/Kyber:
1. For only ML-KEM:
./configure --enable-kyber
./configure --enable-kyber=all,ml-kem
2. For just Kyber:
./configure --enable-kyber=all,original
3. For ML-KEM and Kyber
./configure --enable-kyber=all,original,ml-kem
./configure --enable-kyber=all,ml-kem,original

# Testing
Tested with helium-cli, lightway clients (with and without liboqs)
See: https://polymoon.atlassian.net/browse/CVPN-1554?focusedCommentId=1351477

CI pipelines for `kp_lightway`, `lightway` passed as well

https://github.com/expressvpn/lightway/pull/122


wolfssl-sys/build.rs
```diff
@@ -43,7 +43,10 @@ fn copy_wolfssl(dest: &Path) -> std::io::Result<PathBuf> {
 }
 
 const PATCH_DIR: &str = "patches";
-const PATCHES: &[&str] = &["disable-falcon-dilithium.patch"];
+const PATCHES: &[&str] = &[
+    "disable-falcon-dilithium.patch",
```
xv-raihaan-m, line 47: Do we still need this falcon dilithium patch? We explicitly only enable kyber in the config

wolfssl-sys/patches/use-wolfssl-mlkem-kyber.patch
```diff
@@ -0,0 +1,3325 @@
+diff --git a/IDE/STM32Cube/wolfssl_example.c b/IDE/STM32Cube/wolfssl_example.c
+index 342e8ee9d0..be6195a60a 100644
+--- a/IDE/STM32Cube/wolfssl_example.c
++++ b/IDE/STM32Cube/wolfssl_example.c
```
xv-raihaan-m, line 4: You should preferably have done something closer to

```
git format-patch HEAD~<Number_of_commits>
```

This will split the commits into multiple files so its clear what each patch does and it gives its originating commit.

wolfssl-sys/build.rs
kp-thomas-yau: Seems like we don't. After digging through `configure.ac`, both will only be defined if we enable dilthium and use liboqs. So I will remove this patch

kp-thomas-yau: Actually we still need it lol cuz it turns out it won't store Private key if there's no Falcon or Dilithium... -,-

I think I will also file this bug to a separate ticket to them as well?

kp-thomas-yau: Updated the bug ticket. I think we can merge this if this PR looks good since they are just going to put the update in the new PR anyway?

wolfssl-sys/patches/enable-private-key-for-pqc.patch
kp-mariappan-ramasamy, line 1: Can you please add context around this renaming and removal of the part of the patch ?

wolfssl-sys/build.rs
xv-raihaan-m: I think at this stage its fine to patch it and maybe fix it up later if the future patch looks different

wolfssl-sys/build.rs
```diff
@@ -90,6 +96,8 @@ fn build_wolfssl(wolfssl_src: &Path) -> PathBuf {
         .disable_shared()
         // Disable sys ca certificate store
         .disable("sys-ca-certs", None)
+        // Disable dilithium
+        .disable("dilithium", None)
```
kp-mariappan-ramasamy, line 100: Can you please split this change of disabling dilithium to a separate commit ?

wolfssl-sys/build.rs
kp-mariappan-ramasamy, line 1: Typo in commit message of `[CVPN-1554](https://xv123.atlassian.net/browse/CVPN-1554) Add patch to use WolfSSL Kyber and ML-KEM` :

`maintaing` -> `maintaining`

wolfssl-sys/build.rs
kp-thomas-yau: They have just published a PR for it: https://github.com/wolfSSL/wolfssl/pull/8196
I have replaced the patch with their PR fix

wolfssl-sys/patches/enable-private-key-for-pqc.patch
kp-thomas-yau: I have replaced the patch with their PR fix: https://github.com/wolfSSL/wolfssl/pull/8196

Since now we are removing `liboqs`, `HAVE_LIBOQS` would not be defined and the patch in `settings.h` would not be compiled anyway.

Cargo.lock
kp-mariappan-ramasamy, line 1: Typo in commit message `Remove opq-sys and cargo update `:


`opq-sys` -> `oqs-sys`

wolfssl-sys/Cargo.toml
```diff
@@ -16,21 +16,10 @@ bindgen = "0.70"
...
 [features]
 default = []
 debug = []
-postquantum = ["dep:oqs-sys"]
-
-
-[package.metadata.cargo-all-features]
-# Not an actual feature
-denylist = ["oqs-sys"]
+postquantum = []
```
kp-mariappan-ramasamy, line 22: I think we could enable postquantum as a default feature at this point

kp-mariappan-ramasamy (commented):
Changes LGTM

xv-raihaan-m (approved):
LGTM

kp-mariappan-ramasamy (approved):

PR #195, opened 2024-11-18
CVPN-1445 Add ML-KEM group

kp-thomas-yau:
Add relevant tests and different ML-KEM groups to be used in the Lightway.

The config flags can be found here: https://github.com/expressvpn/wolfssl-rs/pull/192


wolfssl/src/lib.rs
```diff
@@ -236,6 +236,16 @@ pub enum CurveGroup {
...
+
+    /// `WOLFSSL_P256_ML_KEM_512`
+    #[cfg(feature = "postquantum")]
+    P256MLKEMLevel1,
+    /// `WOLFSSL_P384_ML_KEM_768`
+    #[cfg(feature = "postquantum")]
+    P384MLKEMLevel3,
+    /// `WOLFSSL_P521_ML_KEM_1024`
+    #[cfg(feature = "postquantum")]
+    P521MLKEMLevel5,
```
xv-raihaan-m, line 248: This should probably be named the way it is in WolfSSL

xv-raihaan-m:
Probably best to add the usage PR as well so we know this does exactly what we want it to

kp-thomas-yau:
> Probably best to add the usage PR as well so we know this does exactly what we want it to

I am not sure what you mean here?

xv-raihaan-m:
> > Probably best to add the usage PR as well so we know this does exactly what we want it to
> 
> I am not sure what you mean here? @xv-raihaan-m

Adding the PR in expressvpn/lightway so we know that the ML-KEM workflow works.

wolfssl-sys/src/lib.rs
```diff
@@ -40,5 +41,27 @@ mod tests {
             // Check that Kyber was enabled
             assert_eq!(res, WOLFSSL_SUCCESS as c_int);
         }
+
+        // Test ML-KEM is available
```
kp-mariappan-ramasamy, line 45: You can use test-case instead of repeating the test case.

Example:
https://github.com/expressvpn/wolfssl-rs/blob/7a1e3d1933541d7a34009d24a9e4c605296ac6e3/wolfssl/tests/ping_pong.rs#L247

kp-thomas-yau:
> > > Probably best to add the usage PR as well so we know this does exactly what we want it to
> > 
> > 
> > I am not sure what you mean here? @xv-raihaan-m
> 
> Adding the PR in expressvpn/lightway so we know that the ML-KEM workflow works.
See:
https://github.com/expressvpn/lightway/pull/127

I have also tested running lightway client and server locally and it works. See the description in the PR that I have linked here~

kp-mariappan-ramasamy (dismissed):

wolfssl/src/lib.rs
```diff
@@ -236,6 +236,16 @@ pub enum CurveGroup {
     /// `WOLFSSL_P521_KYBER_LEVEL5`
     #[cfg(feature = "postquantum")]
     P521KyberLevel5,
+
+    /// `WOLFSSL_P256_ML_KEM_512`
+    #[cfg(all(feature = "postquantum", not(feature = "kyber_only")))]
```
kp-mariappan-ramasamy, line 241: `kyber_only` uses `postquantum` underneath.

`#[cfg(feature = "postquantum")]` should be enough

kp-thomas-yau: I think if we use `#[cfg(feature = "postquantum")]`, the ML-KEM groups would still be included when we're using `kyber_only`, which is not we want when we are building a kyber-only client

kp-thomas-yau: can I get an approval if this looks fine? @kp-mariappan-ramasamy

xv-raihaan-m (approved):

PR #225, opened 2025-03-03
wolfssl-sys: provide library paths to downstream crates, allow enabling some optional features

pulsastrix:
This PR makes two changes to wolfssl-sys that allow downstream crates to...

A) ...determine the path of the the linked version of wolfSSL. This allows downstream crates that themselves link against C libraries to build and link those against the same version of wolfSSL that is used by this crate (avoiding linking to multiple instances of wolfSSL at once).

To do so, the build script now prints some extra metadata which is then passed to dependent crates by cargo, see https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate. 

B) ...enable some optional features in wolfSSL. While those are not required for wolfssl-rs, dependent C libraries may need them to be available. For now, I have limited this to features that are required for my own use case (see below), but this could of course also be expanded to all other optional features if necessary.

An example use case for the changes in this PR would be linking the `libcoap` binding crate `libcoap-rs` to this crate's version of wolfSSL: namib-project/libcoap-rs#45


kp-mariappan-ramasamy (commented):
Changes look good. Thanks for the contribution.

kp-mariappan-ramasamy:
Our CI is configured to use Earthly satellites, which will not be enabled for forked projects. 
Allow me some time, I will make CI work with forked repos.

kp-mariappan-ramasamy:
I have merged a fix to run CI in github runners instead of earthly satellite. Can you please rebase your branch on top of main ?

pulsastrix:
Done. I also ran `cargo fmt` on `build.rs`, which I forgot to do yesterday.

Thanks for the quick reply!

pulsastrix:
Apparently, the `hmac` feature was already enabled by default in the wolfssl build, and adding it as a separate feature caused it to be disabled when building `wolfssl-sys` for the `wolfssl` crate, my bad.

To fix this (and avoid disabling a previously enabled-by-default feature), I've removed the optional feature, and it should now always be enabled again.

kp-mariappan-ramasamy:
@pulsastrix There is one more fix needed for enabling external contributors.
I have merged those changes in this PR#284 today

Can you please rebase once again ?

kp-mariappan-ramasamy (approved):
LGTM

PR #227, opened 2025-03-04
Add method to generate a collection of random bytes

### LINKED ISSUE #209: Feature request: Add a method to generate a collection of random bytes
I didn't see an obvious way to generate a collection of random bytes so I wrote a quick method to do so in `rng.rs`:

```rust
/// Generates a random collection of bytes. Advances the internal state of the RNG.
pub fn random_bytes<const N: usize>(&mut self) -> Result<[u8; N]> {
    let num_bytes: u32 = N
        .try_into()
        .expect("Failed while trying to convert usize to u32");

    let mut buf = [0u8; N];
    match unsafe {
        wolfssl_sys::wc_RNG_GenerateBlock(&mut self.0 as *mut _, buf.as_mut_ptr(), num_bytes)
    } {
        0 => Ok(buf),
        e => Err(Error::fatal(e)),
    }
}
```

and used it like this which worked fine:
```rust
use wolfssl::Random;

fn main() {
    let mut rand = Random::new().unwrap();
    let bytes = rand.random_bytes::<32>().unwrap();
    println!("{bytes:?}");
}
```

Would the maintainers consider adding a method like this to `rng`?

rmg-x:
Also change `random_u64()` to use this new method internally :^)

Closes #209


kp-mariappan-ramasamy (approved):

kp-mariappan-ramasamy:
@rmg-x We allow only signed commits.
Can you please sign the commit ?

rmg-x:
Signed.

PR #247, opened 2025-04-13
Pre-shared key (PSK)

### LINKED ISSUE #237: Support TLS pre-shared keys
This can be useful when wanting to encrypt a connection without setting up a CA. I think it should be as simple as adding `--enable-psk` to the configure options and then adding wrappers for the related functions https://www.wolfssl.com/using-pre-shared-keys-psk-with-wolfssl-2/

markasoftware:
This adds support for custom user PSKs. Example usage (from added test in `ssl.rs`):

```rust
let client_ctx = ContextBuilder::new(client_method)
    .unwrap_or_else(|e| panic!("new({client_method:?}): {e}"))
    .with_pre_shared_key(&[1, 2, 3, 4])
    .with_secure_renegotiation()
    .unwrap()
    .build();
```

Then do the same on the server side and you're off to the races!

One tradeoff/simplification I made: In general, the client can provide a "client identity" string to the server which is transmitted in the ClientHello, and the server can choose which PSK it uses based on the client identity. That functionality is not supported by this PR, which only allows specifying a single fixed PSK buffer on both client and server.

To support the general case, the wolfSSL api, like the OpenSSL api, doesn't take a fixed buffer but rather takes a callback function that writes the PSK into a mutable buffer. I created global callback functions, and then use `wolfSSL_{get,set}_psk_callback_ctx` to pass the PSK buffer into the callback. This means we have to own the memory for the PSK buffer since we can only store a pointer to it in the `WOLFSSL` object, so I added the PSK buffer as an extra field on the context and session.

I'm a C++ dev by trade and this is my first time contributing to a Rust project so I'm happy to take advice!

Resolves #237


markasoftware:
Just updated this significantly, to allow setting custom server/client callbacks in addition to fixed PSK. Happy to answer questions on the implementation, it's a bit hairy in parts

markasoftware:
still interested in getting this merged, i'm using it actively in my project https://github.com/markasoftware/i405-tunnel

wolfssl/Cargo.toml
```diff
@@ -29,7 +29,7 @@ unnecessary_safety_comment = "deny"
 bytes = "1"
 log = "0.4"
 thiserror = "2.0"
-wolfssl-sys = { path = "../wolfssl-sys", version = "4.0.0" }
+wolfssl-sys = { path = "../wolfssl-sys", version = "4.0.0", features = ["psk"] }
```
kp-mariappan-ramasamy: Can we make this an optional feature too similar to system_ca_certs ?

wolfssl/src/context.rs
```diff
@@ -555,6 +713,73 @@ impl Drop for Context {
...
+pub struct PreSharedKeyClientCallbackResult {
+    /// Should be an empty string if you don't need multiple identities. Else, an arbitrary string
+    /// that the server will be able to read to determine which PSK to use.
+    pub identity: CString,
+    /// The pre-shared key itself.
+    pub key: Vec<u8>,
+}
+
+/// Callbacks that are used to provide a pre-shared key to wolfSSL.
+pub trait PreSharedKeyCallbacks: Debug {
```
kp-mariappan-ramasamy:
```suggestion
pub trait PreSharedKeyCallbacks: Debug + Send + Sync {
```

This drops Send/Sync from Context and Session.

Arc<T>: Send requires T: Send + Sync, and dyn PreSharedKeyCallbacks has only a Debug. So the new Option<Arc<Box<dyn PreSharedKeyCallbacks>>> field on Context (context.rs:673) and Session (ssl.rs:180) makes both types lose the auto impls that `unsafe impl Send for ContextPointer` (context.rs:636), `unsafe impl Sync for ContextPointer` (context.rs:643) and `unsafe impl Send for WolfsslPointer` (context.rs:667) exist to provide.

```
error[E0277]: `(dyn PreSharedKeyCallbacks + 'static)` cannot be shared between threads safely
   = help: the trait `Sync` is not implemented for `(dyn PreSharedKeyCallbacks + 'static)`
   = note: required for `Arc<Box<(dyn PreSharedKeyCallbacks + 'static)>>` to implement `Send`
note: required because it appears within the type `Context`
```

kp-mariappan-ramasamy: Also noticing there are no tests to catch this. Would you mind adding this

```rust
#[allow(dead_code)]
fn _assert_auto_traits<IOCB: IOCallbacks + Send>() {
    fn is_send<T: Send>() {}
    fn is_sync<T: Sync>() {}
    is_send::<Context>();
    is_sync::<Context>();
    is_send::<Session<IOCB>>();
}
```

wolfssl/src/ssl.rs
```diff
@@ -1457,6 +1504,31 @@ mod tests {
...
+                key: PSK.into(),
+            })
+        }
+
+        fn psk_server_callback(
+            &self,
+            identity: &std::ffi::CStr,
+            _max_key_length: usize,
+        ) -> Option<Vec<u8>> {
+            assert!(identity == PSK_IDENTITY);
```
kp-mariappan-ramasamy:
```suggestion
            if identity == PSK_IDENTITY {
                return None;
            }
```

wolfssl/src/context.rs
```diff
@@ -555,6 +713,73 @@ impl Drop for Context {
...
+/// Returned from the client callback in [PreSharedKeyCallbacks]
+pub struct PreSharedKeyClientCallbackResult {
+    /// Should be an empty string if you don't need multiple identities. Else, an arbitrary string
+    /// that the server will be able to read to determine which PSK to use.
+    pub identity: CString,
+    /// The pre-shared key itself.
+    pub key: Vec<u8>,
+}
+
+/// Callbacks that are used to provide a pre-shared key to wolfSSL.
```
kp-mariappan-ramasamy:
```suggestion
 /// Callbacks that are used to provide a pre-shared key to wolfSSL.
///
/// Implementations must not panic. wolfSSL invokes these across an FFI boundary, so a
/// panic is caught and converted into a handshake failure with no detail preserved.
/// Return `None` to reject instead.
```

PR #249, opened 2025-04-16
Lower max MTU for DTLS 1.3 handshake messages to prevent fragmentation with large certificates

xv-pete-m:
#### Context

During the DTLS 1.3 handshake, large post-quantum certificates are exchanged. While DTLS supports message fragmentation, the first packet in the exchange remains extremely close to the MTU limit—just 15 bytes of headroom in the worst-case scenario (e.g., PPPoE or PPPoA networks). This minimal buffer leads to fragmentation on many networks despite the use of handshake fragmentation.

We observed that the current fragmentation behavior splits packets in a ~95/5 ratio, resulting in a large first packet and a tiny second one. This causes the first packet to be fragmented anyway, defeating the purpose of the feature.

#### Change

This patch modifies the `Dtls13SendFragmentedInternal()` function in `dtls13.c` to reduce the maximum fragment size by 500 bytes:

```c
maxFragment = wolfSSL_GetMaxFragSize(ssl, MAX_RECORD_SIZE) - 500;
```

This shifts the packet split to roughly a 70/30 ratio, reducing the likelihood of fragmentation across common network types.

#### Notes

- This change is limited in scope and only affects the initial DTLS 1.3 handshake.
- It is not intended as a permanent fix but rather a practical workaround while a more robust solution is discussed with wolfSSL upstream.

#### Testing

Confirmed that handshake messages now split more evenly, and initial packets no longer exceed common MTU thresholds, preventing fragmentation during connection setup.

### DIFF
--- a/wolfssl-sys/build.rs
+++ b/wolfssl-sys/build.rs
@@ -51,6 +51,7 @@ const PATCHES: &[&str] = &[
     "fix-mlkem-get-curve-name.patch",
     "fix-kyber-get-curve-name.patch",
     "fix-kyber-prf-non-avx2.patch",
+    "CVPN-1945-Lower-max-mtu-for-DTLS-1.3-handshake-message.patch",
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


kp-mariappan-ramasamy (approved):
LGTM

Changes seems to affect only DTLS handshake messages and not Data packets based on callstack of the `Dtls13SendFragmentedInternal`

PR #253, opened 2025-04-29
Add apis for Aes256Gcm cipher

kp-mariappan-ramasamy:
Add apis for Aes256Gcm cipher which can be used by applications.


xv-pete-m (approved):
lgtm

PR #257, opened 2025-05-15
`try_read_slice` and `try_write_slice`

### LINKED ISSUE #256: Does `try_write` really need `data_in: &mut BytesMut`?
Reading the [source code](https://docs.rs/wolfssl/latest/src/wolfssl/ssl.rs.html#468-501) all it does is send the `data_in` to `wolfSSL_write`, which [is documented](https://www.wolfssl.com/documentation/manuals/wolfssl/group__IO.html#function-wolfssl_write) as taking a `const void *`. Does this mean we could just pass a `&Bytes` instead? Or even just a `&[u8]`?

markasoftware:
This is useful for those of us not using `BytesMut` in our own applications. Since `BytesMut` is always on the heap this is good when you want to avoid heap allocations.

I should probably add a test for `try_read_slice` at least since it's not called from `try_read`...

Resolves #256

### DIFF
--- a/wolfssl/src/ssl.rs
+++ b/wolfssl/src/ssl.rs
@@ -466,6 +466,17 @@ impl<IOCB: IOCallbacks> Session<IOCB> {
     // update encryption keys). This can be seen in
     // [`Self::trigger_update_keys`].
     pub fn try_write(&mut self, data_in: &mut BytesMut) -> PollResult<usize> {
+        match self.try_write_slice(data_in) {
+            res @ Ok(Poll::Ready(len)) => {
+                data_in.advance(len);
+                res
+            }
+            other => other,
+        }
+    }
+
+    /// Like try_write, but reads from a simple slice instead.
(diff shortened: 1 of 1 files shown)


kp-mariappan-ramasamy (commented):
LGTM
Minor comment about safety comments.

Also we allow only signed commits to the repo. Can you please sign the commit and push ?

markasoftware:
I've fixed the comment and signed the commit. Thanks!

kp-mariappan-ramasamy (dismissed):

kp-mariappan-ramasamy (approved):

kp-mariappan-ramasamy:
LGTM, thanks for the contribution

markasoftware:
Thanks for the review @kp-mariappan-ramasamy , but I'm unable to merge this:

> New changes require approval from someone other than kp-mariappan-ramasamy because they were the last pusher.

xv-pete-m (approved):

PR #266, opened 2025-06-09
CVPN-2042: Add RISCV-64 support

kp-ben-hung:
Add support for RISCV-64
Fix errors when compiling on RISCV-64, enable RISCV-64 acceleration, and add CI jobs for compiling and tests on RISCV-64 via qemu.

### DIFF
--- a/wolfssl-sys/build.rs
+++ b/wolfssl-sys/build.rs
@@ -193,6 +193,8 @@ fn build_wolfssl(wolfssl_src: &Path) -> PathBuf {
             conf.enable("riscv-asm", None);
             // Disable sp asm optmisations on RISC-V
             conf.disable("sp-asm", None);
+            // Stop frame pointer s0 in RISC-V from being contested
+            conf.cflag("-fomit-frame-pointer");
         }
         _ => {}
     }


.github/workflows/ci.yaml
```diff
@@ -17,13 +17,37 @@ env:
...
+    steps:
+      - uses: earthly/actions-setup@v1
+        with:
+          version: v0.8.3
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Run build command targeting RISCV-64
+        run: earthly --ci +test-riscv64
```
xv-thomas-leong, line 50: Can you put them into `earthly.strategy.matrix.target`? Github Action will pick them up automatically parallelize running all of the targets.

Earthfile
```diff
@@ -74,6 +74,32 @@ build:
     BUILD +run-tests
     BUILD +build-release
 
+build-riscv64:
+    COPY --keep-ts --dir Cargo.toml Cargo.lock deny.toml wolfssl wolfssl-sys ./
```
xv-thomas-leong, line 78: Please reuse the existing `copy-src` target if possible. Do the same for the test target too.

Earthfile
```diff
@@ -74,6 +74,32 @@ build:
...
+    
+    # Install necessary targets and build tools for cross-compiling to RISC-V64 binaries
+    RUN apt-get update -qq && \
+        rustup target add riscv64gc-unknown-linux-gnu && \
+        apt install -y gcc-riscv64-linux-gnu build-essential autoconf autotools-dev libtool-bin clang cmake
```
xv-thomas-leong, line 83: Please perform the build-tools installation in the `build-deps` target, and base your current target on it. Do the same for the test target too.

Earthfile
```diff
@@ -74,6 +74,32 @@ build:
...
+    
+    RUN cargo build --release --target=riscv64gc-unknown-linux-gnu
```
xv-thomas-leong, line 85: Please use the `DO lib-rust+CARGO` earthly functions instead, such that it can utilize caches properly. Do the same for the test target too.

Earthly has a nice blog post of what this function does: https://earthly.dev/blog/incremental-rust-builds/

wolfssl-sys/build.rs
```diff
@@ -188,6 +188,16 @@ fn build_wolfssl(wolfssl_src: &Path) -> PathBuf {
                 conf.enable("aesni", None);
             }
         }
+        // Crate `build-target` only considers riscv32, not v64.
+        // Therefore `riscv64` will match to `Arch::Other`, not `Arch::RISCV`
+        build_target::Arch::Other(arch) => {
```
xv-thomas-leong, line 193: Why is the `RISCV64` symbol not used? You've updated the build-target crate https://github.com/OpenByteDev/build-target/pull/3 to get it to support this symbol, so please use it here.

Earthfile
```diff
@@ -75,30 +75,27 @@ build:
...
 build-riscv64:
-    COPY --keep-ts --dir Cargo.toml Cargo.lock deny.toml wolfssl wolfssl-sys ./
+    COPY --keep-ts --dir Cargo.toml Cargo.lock deny.toml wolfssl wolfssl-sys ./.cargo ./
+    DO lib-rust+INIT --keep_fingerprints=true
     
-    # Install necessary targets and build tools for cross-compiling to RISC-V64 binaries
     RUN apt-get update -qq && \
         rustup target add riscv64gc-unknown-linux-gnu && \
-        apt install -y gcc-riscv64-linux-gnu build-essential autoconf autotools-dev libtool-bin clang cmake
+        apt-get install -y gcc-riscv64-linux-gnu build-essential autoconf autotools-dev libtool-bin clang cmake
```
xv-thomas-leong: Can you extract the dependencies and init into a shared step?

Also please put the COPY step after the dependency step such that file changes won't invalidate cache.

wolfssl-sys/build.rs
```diff
@@ -193,6 +193,8 @@ fn build_wolfssl(wolfssl_src: &Path) -> PathBuf {
             conf.enable("riscv-asm", None);
             // Disable sp asm optmisations on RISC-V
             conf.disable("sp-asm", None);
+            // Stop frame pointer s0 in RISC-V from being contested
+            conf.cflag("-fomit-frame-pointer");
```
xv-thomas-leong, line 197: Can you give some context as to why this is needed, and what are the side-effects of having this?

xv-thomas-leong (changes requested):
Changes look mostly okay, but please clean up your commit history such that it will be easy to understand in the future. Specifically, 

- **Keep meaningful changes only**: the changes you have added and removed don't have to be in commit history
- **Group logically related changes and add meaningful commit messages**: such that they will be easy to read understand in the future

xv-thomas-leong (approved):

PR #278, opened 2025-08-01
WolfSsl: Aes256gcm handle one crypto instance

kp-mariappan-ramasamy:
Having both encrpytion and decryption aes crpyto instance in a single struct
seems more restrictive. Since encrypt and decrypt apis take &mut self, we cannot do encrypt and decrypt parallely.
This restriction is artificially introduced and unnecessary.

Avoid it by making `Aes256Gcm` struct only to manage one crypto instance.
End user application can manage two instances of `Aes256Gcm` for encrpytion and decryption if required to do parallely.


kp-weiwen-goh (approved):
LGTM

PR #283, opened 2025-08-14
wolfssl: add custom error code for ASN_NO_SIGNER error

kp-mariappan-ramasamy:
This makes it easier for client apps to handle this error easily without parsing wolfssl error code.


wolfssl/src/error.rs
```diff
@@ -57,6 +58,9 @@ impl Error {
 /// Abstraction over WolfSSL errors
 #[derive(Clone, Debug, Error)]
 pub enum ErrorKind {
+    /// CA cert not available to verify certificate
+    #[error("CA cert not available to verify certificate")]
+    CaCertNotAvailable,
```
xv-thomas-leong, line 63: Why is the error renamed? The ASN_NO_SIGNER can mean more than just missing CA cert

kp-mariappan-ramasamy: No, It does not mean anything else, unless their documentation lies.
Instead of being cryptic error, I chose to make it more clear.

Ref: https://www.wolfssl.com/documentation/manuals/wolfssl/appendix06.html#asn_no_signer_e-188

xv-thomas-leong: I may have mixed up with another error. Upon checking wolfssl's code, it appears that's what it means.

kp-mariappan-ramasamy: In general, we should try to follow the documentation and not the source code. Even in safety comments, we give pointers to the WolfSsl's documentation. 
And rely on source code only when documentation is missing or not clear.

PR #284, opened 2025-08-15
Fix release CI failing from forked repos

kp-mariappan-ramasamy:
Until now, we use gh-find-current-pr to find the PR number of thespecified commit. This was needed when the action is triggered from main branch push event, since we do not know the PR number from context.

This also works fine for pull requests from branches created in main repo. But it is failing for forked repos.

Example (logs may be purged in future):
https://github.com/expressvpn/wolfssl-rs/actions/runs/16980030164/job/48137907811

To overcome this issue, use `github.event.number` for pull request event and use the one from above action if it is push event.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,11 +27,19 @@ jobs:
           state: all
       - id: pr_labels
         uses: actions/github-script@v7
+        env:
+          PR_NUMBER: ${{ steps.find_pr.outputs.pr }}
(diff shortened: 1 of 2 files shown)


kp-mariappan-ramasamy:
Verified it by running with a temp commit on Cargo.toml:
https://github.com/expressvpn/wolfssl-rs/actions/runs/16981661097/job/48142647097?pr=284

xv-thomas-leong (approved):

PR #295, opened 2025-09-26
wolfssl-sys: Add Mac catalyst support

kp-said-rehouni:
This PR adds support to Mac Catalyst by updating the build script and CI configuration to compile for this platform.


wolfssl-sys/build.rs
```diff
@@ -485,6 +485,37 @@ fn build_wolfssl(wolfssl_src: &Path) -> PathBuf {
...
+        conf.cxxflag(arch_flags);
+        conf.env("ARCH", arch);
+
+        // General Mac Catalyst specific configurations
+        conf.disable("crypttests", None);
+        conf.cflag("-D_FORTIFY_SOURCE=2");
+        if cfg!(feature = "system_ca_certs") {
+            conf.cflag("-DWOLFSSL_APPLE_NATIVE_CERT_VALIDATION");
+        }
+    }
```
kp-thomas-yau, line 517: The target_os of mac catalyst is also set to `iOS` as well, so I think you might need to move this part under iOS section.

Plus, you can use this to differentiate it from iOS or mac catalyst (from https://doc.rust-lang.org/beta/rustc/platform-support/apple-ios-macabi.html#building-rust-programs):
```rust
if cfg!(target_env = "macabi") {
    // Do something only on Mac Catalyst.
}
```

kp-said-rehouni: Good catch! Thanks a lot @kp-thomas-yau for the feedback.
Just updated it

kp-thomas-yau: Whoops, i have just realised cfg!(target_abi) doesn't work in build scripts during cross-compilation, just updated the code to read env var instead

xv-thomas-leong (approved):

PR #361, opened 2026-05-07
debug: add install_logging_callback for safe Rust callbacks

kp-shawn-thiah:
The existing wolfssl::set_logging_callback exposes directly, forcing every consumer to write the FFI boilerplate  themselves (unsafe extern "C", null check, CStr conversion, UTF-8 fallback catch_unwind for panic safety). Hoist that into wolfssl-rs once.
  - Add an internal trampoline that handles the FFI translation once at                                                                                                                                                      
    the C boundary.                                                                                                                                                                                                          
  - Expose install_logging_callback(cb: fn(&str)) — registers the                                                                                                                                                            
    trampoline with wolfSSL_SetLoggingCb and stashes cb in a OnceLock                                                                                                                                                        
    for the trampoline to dispatch into.                                                                                                                                                                                     
  - Rename the existing FFI-shaped LoggingCallback alias to                                                                                                                                                                  
    RawLoggingCallback, pub(crate). The matching set_logging_callback                                                                                                                                                        
    follows — its parameter type is now opaque externally, so it's                                                                                                                                                           
    effectively crate-internal.                                                                                                                                                                                              
  - Bump wolfssl to 7.0.0 (major) for the breaking rename.

### DIFF
--- a/wolfssl/src/debug.rs
+++ b/wolfssl/src/debug.rs
@@ -1,9 +1,11 @@
 #![cfg(feature = "debug")]
+#![allow(unsafe_code)]
 
+use std::ffi::{c_char, CStr};
 use std::fmt::{self, Display};
 #[allow(unused_imports)] // Needed for windows
 use std::os::raw::{c_int, c_uint};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 
 /// Application provided callbacks to receive TLS1.3 secrets
 ///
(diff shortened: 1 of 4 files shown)


wolfssl/src/debug.rs
```diff
@@ -113,3 +115,34 @@ impl From<c_int> for Tls13Secret {
...
+unsafe extern "C" fn logging_trampoline(_level: c_int, msg: *const c_char) {
+    if msg.is_null() {
+        return;
+    }
+    // SAFETY: null-checked; wolfSSL emits NUL-terminated strings via snprintf.
+    let c_str = unsafe { CStr::from_ptr(msg) };
+    let message = c_str.to_str().unwrap_or("Unable to decode C string");
+
+    if let Some(cb) = LOGGING_CALLBACK.get() {
+        let _ = std::panic::catch_unwind(|| cb(message));
```
kp-mariappan-ramasamy: Guess we could add a warn log here, if it errors out
Otherwise we might have missing logs with no clue

kp-shawn-thiah: sounds good, just added a warn log to saying that the logging callback error-ed.

kp-mariappan-ramasamy (approved):

PR #368, opened 2026-06-16
build: windows: build with 2022 and 2026 msbuild

kp-antonio-yang:


wolfssl-sys/build.rs
```diff
@@ -240,7 +240,7 @@ Builds WolfSSL in windows
 */
 #[cfg(windows)]
 fn build_win(wolfssl_src: &Path) -> PathBuf {
-    let msb = MsBuild::find_msbuild(Some("2022")).expect("Failed to find MsBuild 2022");
+    let msb = MsBuild::find_msbuild(None).expect("Failed to find MsBuild");
```
kp-mariappan-ramasamy: There is an api to choose from different version in Msbuild crate, you can use that:
https://docs.rs/crate/msbuild/0.2.1/source/src/lib.rs#146


So we can support both 2022 and 2026. Though you may to update PlatformToolset https://github.com/expressvpn/wolfssl-rs/pull/368/changes#diff-a9ea68a2baea9d9cdb88b6f3b3275b3b89bfebe8cb9b3d28d8dbc00092bc6d89R254 based on the version installed.

kp-thomas-yau (changes requested):
We should also update the CI machine for Windows to the latest one as well to test whether this change here actually works. Plus, this change should also be released as well so that downstream repo can actually use this change...

kp-mariappan-ramasamy (approved):
LGTM thanks

.github/workflows/ci.yaml
```diff
@@ -256,11 +256,17 @@ jobs:
...
+          - target: x86_64-pc-windows-msvc
+            os: windows-2025  # Windows Server 2025 with Visual Studio 2026
             arch: x64
           - target: i686-pc-windows-msvc
-            os: windows-2022
+            os: windows-2022  # with Visual Studio 2022
             arch: x86
+          - target: i686-pc-windows-msvc
+            os: windows-2025  # Windows Server 2025 with Visual Studio 2026
+            arch: x64
```
kp-thomas-yau, line 269: nit: I think we can just test with windows-latest? So we can know something went wrong on the latest Windows machine

kp-thomas-yau (commented):
LGTM, but you might want to update PR title here lol

kp-thomas-yau (approved):

.github/workflows/ci.yaml
kp-antonio-yang: Good point, I think this can do in down stream.

PR #372, opened 2026-07-14
build: default Apple deployment targets to rustc's deployment target

kp-thomas-yau:
## Why

Every consumer of `wolfssl-sys` on Apple platforms currently has to export `MACOSX_DEPLOYMENT_TARGET` / `IPHONEOS_DEPLOYMENT_TARGET` / `TVOS_DEPLOYMENT_TARGET` or the build panics. The panic was caution against a real hazard: with the variable unset, the C build would silently target whatever cc-rs falls back to — the Xcode SDK version (26.2 on current Xcode) or a hardcoded per-arch value. That fallback is machine-dependent and can exceed the minimum OS the Rust side links against, which means a binary that loads fine in testing but can call APIs that don't exist on older customer devices.

`rustc --print deployment-target` provides a default that is correct by construction: the C objects are built for the same minimum OS version as the Rust objects they are linked with, so the two halves of the binary can't disagree. With a safe default available, the env vars become optional overrides rather than a hard requirement — unchanged for our CI, which sets them explicitly — and an unset var now produces a visible `cargo:warning` naming the chosen value instead of a panic.

## How

- Resolve each deployment target as: env var if set, otherwise rustc's value for the target triple (plus a warning).
- Put the resolved value in the build script's process env before `conf.build()`. Verified empirically that this is the only channel that works: cc-rs reads it there and bakes `-m*-version-min=` into the CFLAGS handed to configure, and that explicit flag beats any env var the configure/make subprocesses see (`Config::env()`-only values never reach the compiled objects; checked via config.log and `otool -l` minos with distinct test values).
- Declare `cargo:rerun-if-env-changed` so changing a deployment target rebuilds the C library — previously this required a manual clean.

## Notes

- macOS was verified end-to-end locally (default, override, and rebuild-on-change paths); iOS/Catalyst/tvOS use the identical mechanism and are exercised by CI.
- `env::set_var` is safe under edition 2021; a move to edition 2024 would need a one-line `unsafe` wrap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -773,7 +773,7 @@ dependencies = [
 
 [[package]]
 name = "wolfssl-sys"
-version = "4.2.1"
+version = "4.3.0"
 dependencies = [
  "autotools",
  "bindgen",
--- a/wolfssl-sys/Cargo.toml
+++ b/wolfssl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wolfssl-sys"
-version = "4.2.1"
+version = "4.3.0"
 description = "System bindings for WolfSSL"
 readme = "README.md"
 authors.workspace = true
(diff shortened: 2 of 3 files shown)


kp-weiwen-goh (approved):
LGTM

PR #386, opened 2026-08-25
wolfssl-sys: Enable Windows x86_64 assembly speedups

kp-thomas-yau:
Enable the x64 MASM assembly in the Windows build. `wolfssl.vcxproj` already
assembles all six MASM files, but none of the macros that make the C code call
them were defined:

- `WOLFSSL_SP_ASM` + `WOLFSSL_SP_X86_64_ASM` — assembly SP math for RSA/ECC
- `USE_INTEL_CHACHA_SPEEDUP` + `USE_INTEL_POLY1305_SPEEDUP` — AVX/AVX2 ChaCha20-Poly1305
- `HAVE_INTEL_AVX1/AVX2` — AVX AES-GCM (previously SSE AESNI only)

All paths dispatch at runtime via cpuid. `USE_INTEL_SPEEDUP` stays off: it pulls
in SHA-2 assembly with no MASM port in wolfSSL v5.9.1; revisit on the next bump.

Included a PR fix from WolfSSL team for link failure when we build SP_ASM with SP_4096/384: wolfSSL/wolfssl#11254

Also in this PR:
- `build.rs` declares `cargo:rerun-if-changed=windows` — user_settings edits previously reused stale libraries silently
- GCM table GHASH variants benchmarked and rejected; numbers recorded in `user_settings-common.h`
- Version bump: wolfssl-sys 4.4.0, wolfssl 7.5.0

Verified: full wolfssl + wolfssl-sys test suites pass on ARM64 Windows (129 tests); full wolfCrypt self-test passes on x64.

## Performance: x64 Windows before/after

wolfCrypt benchmark (`benchmark.exe`, Release x64, MSVC), Intel i7-1165G7.
Medians of 5 interleaved runs per variant, pinned to one core.
**Before** = branch parent (C fallbacks). **After** = this PR (SP/AVX assembly enabled,
`sp_x86_64_asm.asm` link fix from wolfSSL/wolfssl#11254, ml64 defines trimmed to SP_384/SP_4096).

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| RSA-2048 private | 156 ops/s | 1,952 ops/s | **12.5×** |
| RSA-2048 public | 7,970 ops/s | 61,997 ops/s | **7.8×** |
| ECDSA P-256 sign | 6,562 ops/s | 53,465 ops/s | **8.1×** |
| ECDSA P-256 verify | 3,463 ops/s | 18,500 ops/s | **5.3×** |
| ECDHE P-256 agree | 3,577 ops/s | 22,348 ops/s | **6.2×** |
| ECDSA P-384 sign | 2,217 ops/s | 17,466 ops/s | **7.9×** |
| ECDSA P-384 verify | 1,083 ops/s | 5,751 ops/s | **5.3×** |
| ECDHE P-384 agree | 1,208 ops/s | 6,314 ops/s | **5.2×** |
| ChaCha20 | 465 MiB/s | 3,057 MiB/s | **6.6×** |
| Poly1305 | 1,302 MiB/s | 6,634 MiB/s | **5.1×** |
| ChaCha20-Poly1305 AEAD | 341 MiB/s | 2,029 MiB/s | **5.9×** |
| AES-GCM (128/192/256, enc+dec) | 5.0–5.8 GiB/s | 5.6–6.3 GiB/s | +1–15% |

Controls (AES-CBC, SHA-256, GMAC) were flat within run noise (−2.5%…+6%), so the deltas are attributable to the enabled assembly. AES-GCM is the noisiest family: its gain (AVX vs SSE AESNI) is real but small next to the public-key and ChaCha20-Poly1305 wins. Full wolfCrypt self-test passes on both variants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -780,7 +780,7 @@ dependencies = [
 
 [[package]]
 name = "wolfssl"
-version = "7.4.0"
+version = "7.5.0"
 dependencies = [
  "async-trait",
  "bytes",
@@ -795,7 +795,7 @@ dependencies = [
 
 [[package]]
 name = "wolfssl-sys"
-version = "4.3.1"
+version = "4.4.0"
 dependencies = [
  "autotools",
  "bindgen",
--- a/wolfssl-sys/Cargo.toml
+++ b/wolfssl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wolfssl-sys"
-version = "4.3.1"
+version = "4.4.0"
 description = "System bindings for WolfSSL"
 readme = "README.md"
 authors.workspace = true
--- a/wolfssl-sys/build.rs
+++ b/wolfssl-sys/build.rs
@@ -41,6 +41,9 @@ fn copy_wolfssl(dest: &Path) -> std::io::Result<PathBuf> {
     copy_dir_recursive(src, &dest_dir)?;
 
     if build_target::target_os() == build_target::Os::Windows {
+        // The generated user_settings.h is assembled from these headers
+        println!("cargo:rerun-if-changed=windows");
+
         // Determine architecture-specific user_settings file
         let arch_settings = match build_target::target_arch() {
             build_target::Arch::X86_64 => "windows/user_settings-x86_64.h",
--- a/wolfssl-sys/build.rs
+++ b/wolfssl-sys/build.rs
@@ -145,6 +145,7 @@ const PATCH_DIR: &str = "patches";
 const PATCHES: &[&str] = &[
     "CVPN-1945-Lower-max-mtu-for-DTLS-1.3-handshake-message.patch",
     "PR10492-DTLS13-backward-compatible.patch",
+    "PR11254-MSVC-pass-SP-defines-to-ml64.patch",
 ];
 const OPTIONAL_FEATURES: &[&str] = &["aesccm", "dh", "opensslall", "opensslextra", "psk"];
 const MACRO_FEATURES: &[(&str, &str)] = &[("ex_data", "HAVE_EX_DATA"), ("alpn", "HAVE_ALPN")];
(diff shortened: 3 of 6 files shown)


kp-thomas-yau:
For actual Lightway performances:

## Benchmark: baseline vs asm speedup (Windows x64)

**Setup:** Windows 11 x64, 10 GbE NIC, lightway UDP with AES-256-GCM.
Ookla speedtest CLI pinned to server 48446 (GSL Networks, Hong Kong);
10 measured rounds per build + 1 discarded warm-up, 10 s cooldown, identical config and VPN server for both builds. Client process CPU time sampled around each round.

### Results (median of 10 rounds)

| Metric | baseline | asm speedup | Δ |
|---|---|---|---|
| Download (Mbps) | 1321 | 1434 | **+8.5%** |
| Upload (Mbps) | 1054 | 1096 | +4.0% (within noise) |
| Efficiency (Mbit per CPU-s) | 1164 | 1219 | **+4.7%** |
| UL loaded latency (ms) | 73.3 | 50.8 | **−30.7%** |
| UL loaded jitter (ms) | 18.3 | 14.2 | **−22.6%** |
| Idle latency (ms) | 2.41 | 2.43 | unchanged |
| DL loaded latency (ms) | 10.34 | 10.29 | unchanged |

### TLDR

- Both builds ran **client-bound** (~90% of one core for the whole test), so the deltas measure the client itself, not the network path.
- The **download gain is statistically solid**: +113 Mbps against ~40 Mbps stddev (Welch t ≈ 6). The distributions barely overlap — baseline's best round (1406 Mbps) is below the speedup build's median.
- **Upload loaded latency/jitter drop ~23–31%** (~3.5σ): the speedup build drains the send path faster, so less queuing builds up under upload load.
- Upload throughput (+4%, ~1.5σ) is plausible but not confirmed at 10 rounds; idle latency/jitter and download loaded latency are unchanged.
- The speedup build is **still pegged on a single core at ~1.4 Gbps** — each byte got cheaper and the ceiling moved up proportionally. Multi-gigabit rates will need further crypto speedups or optimizations for Wintun.

kp-mariappan-ramasamy (commented):
Changes lgtm

Can we merge this after https://github.com/wolfSSL/wolfssl/pull/11254/changes ?
Otherwise we might need to redo if there are more changes in the upstream pr.

PR #387, opened 2026-08-26
wolfssl-sys: Use the NDK archiver for Android builds

kp-thomas-yau:
Autotools' configure looks for `<host-triplet>-ar`/`-ranlib` and otherwise falls back to the unprefixed host tools. Modern NDKs ship neither, so on a macOS host configure settles for Xcode's `ar`/`ranlib`, which cannot index ELF objects and leaves `libwolfssl.a` a 96-byte empty archive. Nothing fails: the archive is bundled into the rlib, the cdylib links with every `wolfSSL_*` symbol undefined, and the breakage only surfaces at dlopen.

Pass configure the archiver cc-rs resolves, which is the same resolution already used for CC — `AR_<triple>`/`RANLIB_<triple>` when the caller sets them (cargo-ndk does), else the NDK's `llvm-ar`/`llvm-ranlib`.

Linux hosts were unaffected because GNU ranlib indexes ELF fine, which is why CI never caught it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -795,11 +795,12 @@ dependencies = [
 
 [[package]]
 name = "wolfssl-sys"
-version = "4.3.0"
+version = "4.3.1"
 dependencies = [
  "autotools",
  "bindgen",
  "build-target",
+ "cc",
  "msbuild",
  "test-case",
 ]
--- a/wolfssl-sys/Cargo.toml
+++ b/wolfssl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wolfssl-sys"
-version = "4.3.0"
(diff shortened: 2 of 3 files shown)


xv-thomas-leong (approved):

kp-mariappan-ramasamy:
@kp-thomas-yau how was this undetected until now ? 
Some of our team were using macbook for android development, werent they

kp-thomas-yau:
> @kp-thomas-yau how was this undetected until now ? Some of our team were using macbook for android development, werent they

I think newer XCode SDK behaves differently, where older XCode SDK index ELF but newer XCode SDK (I've just bumped to use XCode 26.5) don't.

PR #388, opened 2026-08-26
[CVPN-2620] release flow hardening

xv-piotr-haber:
Harden the release workflow according to checklist:
[x] Move CARGO_REGISTRY_TOKEN out of job-level env; scope it to the single publish step only.

[x] Split pipeline: no-secret planning job (npm + TOML parsing) → separate publish job holding the token.

[x] Replace ad-hoc npm install with committed package.json + lockfile and npm ci; pin exact versions, or vendor the helper logic so nothing is resolved at release time.

[x] Use npm ci --ignore-scripts where lifecycle scripts aren't needed.

[x] Set persist-credentials: false on actions/checkout unless a later git-authenticated step needs it.

[x] Pin GitHub Actions by commit SHA; add minimal permissions: blocks.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -100,6 +100,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: true
+          persist-credentials: false
 
       - name: Check Cargo.lock is up-to-date
         shell: bash
@@ -318,6 +319,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: true
+          persist-credentials: false
 
       - uses: actions/setup-node@v4
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,19 +11,24 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.ref_name != 'main' }}
 
+permissions: {}
+
 jobs:
   check_label:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+      issues: read
     outputs:
       is_ignore_release_labelled: ${{ steps.pr_labels.outputs.is_ignore_release_labelled }}
       issue_number: ${{ steps.pr_labels.outputs.issue_number }}
     steps:
-      - uses: jwalton/gh-find-current-pr@v1
+      - uses: jwalton/gh-find-current-pr@f3d61b485d2801773f7a07b2aaa3306bd8f8e653 # v1.3.5
         id: find_pr
         with:
           state: all
       - id: pr_labels
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
         env:
           PR_NUMBER: ${{ steps.find_pr.outputs.pr }}
           EVENT_NUMBER: ${{ github.event.number }}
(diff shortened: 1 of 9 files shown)


xv-piotr-haber:
Verified dry-run release step in separate [PR](https://github.com/expressvpn/wolfssl-rs/pull/389) that bumps wolfssl version

.github/workflows/release.yaml
```diff
@@ -88,18 +88,15 @@ jobs:
             :label: Release skipped — `ignore-release` label is applied.
           edit-mode: replace
 
-  release:
+  plan:
```
xv-thomas-leong, line 99: Do we need a plan job here? Would a separate release step be sufficient?

xv-thomas-leong (commented):
Some comments about commit hygiene:

- For commit 75d1b8b, can we separate the formatting and the actual change?
- Can the commit aacd5fd be folded into the one before?

.github/workflows/release.yaml
xv-piotr-haber: Not sure what you mean.
I've split the `release` job into `plan` and `publish` jobs as requested.
I don't think job names matter that much but these new describe what is being done quite good.
For extra clarity we might name them `plan-release` and `publish-release`

Or do you mean `plan` should not be a separate job but just a step.
I guess we could put it in `check-label` job  but that might be a bit confusing

xv-thomas-leong: > Or do you mean plan should not be a separate job but just a step.

Yes, I mean a step should be sufficient

xv-piotr-haber: Not sure i can do that without making it confusing.
`plan` job consists of multiple steps,
we could move those steps to `check_label` job
but then we would also need to integrate steps done in `skip_release` job as well for consistency.
We could also move those `plan` steps to `publish` job but the explicit ask was to separate those two.

xv-piotr-haber: I folded 'check_label' + 'skip_release' + 'plan' jobs into one named 'plan'

.github/workflows/release.yaml
```diff
@@ -359,23 +289,161 @@ jobs:
...
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '20.x'
+      - run: npm ci --ignore-scripts
+
+      - name: Publish crates sequentially
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        id: process_release
+        env:
+          CARGO_REGISTRY_TOKEN: "${{ secrets.CARGO_REGISTRY_TOKEN }}"
```
xv-thomas-leong: We can further restrict the token to only be available in non-dry-run mode.

```suggestion
          CARGO_REGISTRY_TOKEN: ${{ needs.plan.outputs.is_dry_run == 'false' && secrets.CARGO_REGISTRY_TOKEN || '' }}
```

.github/workflows/release.yaml
```diff
@@ -359,23 +289,161 @@ jobs:
...
+          IS_DRY_RUN: ${{ needs.plan.outputs.is_dry_run }}
+        with:
+          script: |
+            const axios = require('axios');
+            const { spawn } = require('child_process');
+
+            const is_dry_run = process.env.IS_DRY_RUN === 'true';
+            const publish_plan = JSON.parse(process.env.PUBLISH_PLAN || '[]');
+
+            async function get_crate_max_version(name) {
```
xv-thomas-leong: These functions are being duplicated for the third time. Can we extract it into a js file and keep them DRY?

https://github.com/actions/github-script#run-a-separate-file-with-an-async-function

xv-thomas-leong (commented):
Apologies for more comments, this hardening change necessitates a bit of refactoring

.github/workflows/release.yaml
xv-piotr-haber: done

.github/workflows/release.yaml
xv-piotr-haber: done

xv-thomas-leong (approved):

